PR #7, opened 2023-05-04
PoC of Tauri working with `react-resizable-panels`

sergioramos:


hammer:
@sergioramos do you want a review or should we just commit?

Issue #4, opened 2023-05-03
[Discussion] Text editing libraries


hammer:
TipTap vs. Lexical.

TipTap the default choice.

3 side quests to evaluate:
1. Text/block folding
2. Fine-grained customizability in text editing
3. Collaboration

hammer:
For 2, @danvk found https://lexical.dev/docs/demos/plugins/collapsible-container for Lexical

hammer:
For 3, we are most concerned with batch, asynchronous interactions similar to pull requests. We are not concerned about synchronous, "real-time" collaboration.

hammer:
Beyond the 3 side quests mentioned above, as we tend towards the IDE UX, we'll need to care about having multiple editors open at once, so behavior in that scenario should probably be considered as well.

hammer:
We're pretty far down the road with TipTap so I'm going to close this for now and we can revisit if we hit any showstopper issues that might make us consider Lexical again.

PR #13, opened 2023-05-16
POC using Lexical Editor

cguedes:
This PR is the POC for using Lexical within the evaluation of text editors being discussed in #4 

POC with 3 panels:
 - Left Panel: References list
 - Center Panel: Lexical editor with rich-text
 - Right Panel : AI interactions

Selections of text are listed in the right panel, and clicks on a reference (left panel) add the text (ex: Ref 3) to the selected/cursor area in the editor (central panel).

image

PR #16, opened 2023-05-18
Poc editor tiptap

sehyod:
This PR is the PoC for using TipTap within the evaluation of text editors being discussed in #4 

POC with 3 panels:

Left Panel: List of project's documents
Center Panel: TipTap editor
Right Panel : references and AI interactions
- Selecting text in the editor send it to the backend and shows the result in the AI interactions panel
- Clicking on a reference adds it to the text as a special, uneditable node

The editor supports basic text formatting and draggable collapsible blocks.

Due to tauri's limitation (see this [issue](https://github.com/tauri-apps/tauri/issues/3277)), HTML drag events don't work properly with the `fileDropEnabled` config

### DIFF
--- a/src/CollapsibleBlock/CollapsibleBlock.tsx
+++ b/src/CollapsibleBlock/CollapsibleBlock.tsx
@@ -1,0 +1,26 @@
+import { NodeViewContent, NodeViewWrapper } from '@tiptap/react';
+import './styles.css';
+
+export const CollapsibleBlock = (props: any) => {
+    const handleButtonClick = () => {
+        props.updateAttributes({
+            folded: !props.node.attrs.folded,
+        });
+    };
+    return (
+        <NodeViewWrapper>
+            <div className='draggable-item collapsible-block' style={{ display: 'flex', flexDirection: 'row', alignItems: 'flex-start' }}>
+                <div
+                    className="drag-handle"
+                    contentEditable="false"
+                    draggable="true"
+                    data-drag-handle
+                    style={{ backgroundImage: `url('data:image/svg+xml;charset=UTF-8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 16"><path fill-opacity="0.2" d="M4 14c0 1.1-.9 2-2 2s-2-.9-2-2 .9-2 2-2 2 .9 2 2zM2 6C.9 6 0 6.9 0 8s.9 2 2 2 2-.9 2-2-.9-2-2-2zm0-6C.9 0 0 .9 0 2s.9 2 2 2 2-.9 2-2-.9-2-2-2zm6 4c1.1 0 2-.9 2-2s-.9-2-2-2-2 .9-2 2 .9 2 2 2zm0 2c-1.1 0-2 .9-2 2s.9 2 2 2 2-.9 2-2-.9-2-2-2zm0 6c-1.1 0-2 .9-2 2s.9 2 2 2 2-.9 2-2-.9-2-2-2z" /></svg>')` }}
+                />
+                <button className={props.node.attrs.folded ? '' : 'open'} onClick={handleButtonClick} />
+
+                <NodeViewContent className={`content ${props.node.attrs.folded ? "folded" : ""}`} />
+            </div>
+        </NodeViewWrapper>
+    )
+}
--- a/src/CollapsibleBlock/nodes/CollapsibleBlock.ts
+++ b/src/CollapsibleBlock/nodes/CollapsibleBlock.ts
@@ -1,0 +1,45 @@
+import { mergeAttributes, Node } from '@tiptap/core';
+import { ReactNodeViewRenderer } from '@tiptap/react';
+import { CollapsibleBlock } from '../CollapsibleBlock';
+
+
+export const CollapsibleBlockNode = Node.create({
+  name: 'collapsibleBlock',
+
+  group: 'block',
+
+  content: 'collapsibleSummary collapsibleContent?',
+  draggable:true,
+
+  parseHTML() {
+    return [
+      {
+        tag: 'collapsible-block',
+      },
+    ]
+  },
+
+  addKeyboardShortcuts() {
+    return {
+      'Mod-Enter': () => {
+        return this.editor.chain().insertContentAt(this.editor.state.selection.head, { type: this.type.name }).focus().run()
+      },
+    }
+  },
+
+  renderHTML({ HTMLAttributes }) {
+    return ['collapsible-block', mergeAttributes(HTMLAttributes), 0]
+  },
+
+  addAttributes() {
+    return {
+      folded: {
+        default: true,
+      }
+    }
+  },
+
+  addNodeView() {
+    return ReactNodeViewRenderer(CollapsibleBlock)
+  },
+})
--- a/src/Reference/ReferencesList.tsx
+++ b/src/Reference/ReferencesList.tsx
@@ -1,0 +1,70 @@
+import './ReferencesList.css'
+
+import {
+    forwardRef, useEffect, useImperativeHandle,
+    useState,
+} from 'react'
+
+export default forwardRef((props: { items: Array<string>, command: ({ id }: { id: string }) => any }, ref) => {
+    const [selectedIndex, setSelectedIndex] = useState(0)
+
+    const selectItem = (index: number) => {
+        const item = props.items[index]
+
+        if (item) {
+            props.command({ id: item })
+        }
+    }
+
+    const upHandler = () => {
+        setSelectedIndex((selectedIndex + props.items.length - 1) % props.items.length)
+    }
+
+    const downHandler = () => {
+        setSelectedIndex((selectedIndex + 1) % props.items.length)
+    }
+
+    const enterHandler = () => {
+        selectItem(selectedIndex)
+    }
+
+    useEffect(() => setSelectedIndex(0), [props.items])
+
+    useImperativeHandle(ref, () => ({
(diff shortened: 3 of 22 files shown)


hammer:
Nice! A lot of the comments from https://github.com/refstudio/refstudio/pull/13#issuecomment-1553256104 also apply here.

A few additional thoughts
- It's nice to see blocks with handles that can be relocated here.
- Text selections showing up in the AI interactions component already! I didn't notice this on the Lexical demo, I think it's working there too.
- Toggling block collapse with Cmd-Enter does not work as expected.
- File browser +/- not yet working it seems
- Vertical scroll within a component not working yet

hammer:
@sergioramos @danvk @cguedes could y'all review this PR and get it ready for the merge to `main`?

src/CollapsibleBlock/nodes/CollapsibleBlock.ts
danvk, line 1: Make the file name match the class name: `CollapsibleBlockNode.ts`. (Similar comment for the next few files.)

src/CollapsibleBlock/CollapsibleBlock.tsx
```diff
@@ -0,0 +1,26 @@
+import { NodeViewContent, NodeViewWrapper } from '@tiptap/react';
+import './styles.css';
+
+export const CollapsibleBlock = (props: any) => {
```
danvk, line 4: Define a TypeScript type for props.

src/CollapsibleBlock/CollapsibleBlock.tsx
```diff
@@ -0,0 +1,26 @@
...
+    };
+    return (
+        <NodeViewWrapper>
+            <div className='draggable-item collapsible-block' style={{ display: 'flex', flexDirection: 'row', alignItems: 'flex-start' }}>
+                <div
+                    className="drag-handle"
+                    contentEditable="false"
+                    draggable="true"
+                    data-drag-handle
+                    style={{ backgroundImage: `url('data:image/svg+xml;charset=UTF-8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 16"><path fill-opacity="0.2" d="M4 14c0 1.1-.9 2-2 2s-2-.9-2-2 .9-2 2-2 2 .9 2 2zM2 6C.9 6 0 6.9 0 8s.9 2 2 2 2-.9 2-2-.9-2-2-2zm0-6C.9 0 0 .9 0 2s.9 2 2 2 2-.9 2-2-.9-2-2-2zm6 4c1.1 0 2-.9 2-2s-.9-2-2-2-2 .9-2 2 .9 2 2 2zm0 2c-1.1 0-2 .9-2 2s.9 2 2 2 2-.9 2-2-.9-2-2-2zm0 6c-1.1 0-2 .9-2 2s.9 2 2 2 2-.9 2-2-.9-2-2-2z" /></svg>')` }}
```
danvk, line 18: Factor this SVG string out into a variable that explains what it is, e.g.

```ts
const DRAG_HANDLE_SVG = `url(...)`;
```

or put it in an SVG file and `import` that (it should be possible to get this to work with both vite and TypeScript).

src/CollapsibleBlock/CollapsibleBlock.tsx
```diff
@@ -0,0 +1,26 @@
...
+                />
+                <button className={props.node.attrs.folded ? '' : 'open'} onClick={handleButtonClick} />
+
+                <NodeViewContent className={`content ${props.node.attrs.folded ? "folded" : ""}`} />
```
danvk, line 22: Suggest using the `classNames` library for `className` construction.

src/Reference/ReferencesList.tsx
```diff
@@ -0,0 +1,70 @@
+import './ReferencesList.css'
+
+import {
+    forwardRef, useEffect, useImperativeHandle,
+    useState,
+} from 'react'
+
+export default forwardRef((props: { items: Array<string>, command: ({ id }: { id: string }) => any }, ref) => {
```
danvk, line 8: pull the Props type declaration out into an interface

danvk, line 8: use `=> void` instead of `=> any` if you don't care about the return type of a callback.

src/Reference/ReferencesList.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const [selectedIndex, setSelectedIndex] = useState(0)
+
+    const selectItem = (index: number) => {
+        const item = props.items[index]
```
danvk, line 12: Suggest using destructuring assignment to avoid repeating `props.` throughout this component:

```ts
const {items, command} = props;
```

src/types/ReferenceItem.ts
```diff
@@ -0,0 +1,7 @@
+export type ReferenceItem = {
```
danvk, line 1: Suggest using `interface` for simple types, see #18

src/types/ReferenceItem.ts
```diff
@@ -0,0 +1,7 @@
...
+    id: string;
+    title: string;
+    author: string;
+    year: string;
```
danvk, line 5: `number`?

src/views/AIView.tsx
```diff
@@ -0,0 +1,35 @@
...
+                marginTop: "1rem",
+                marginBottom: "1rem",
+                fontStyle: 'italic'
+            }}>
+                Select some text in the editor to see it here.
+            </p>
+
+            {selection && (
+                <div style={{ display: 'flex', flexDirection: "column", gap: "1rem" }}>
+                    <div style={{ borderWidth: "1px", borderColor: "rgb(254 249 195)", backgroundColor: "rgb(254 252 232)", padding: "1rem" }}>
```
danvk, line 26: TIL that you don't need to put commas in `rgb`! https://css-tricks.com/no-comma-color-functions-in-css/

src/views/AIView.tsx
```diff
@@ -0,0 +1,35 @@
+import { invoke } from '@tauri-apps/api/tauri';
+import { useEffect, useState } from 'react';
+
+export function AIView({ selection }: { selection: string }) {
+    const [reply, setReply] = useState('');
+
+    useEffect(() => {
+        invoke<string>('interact_with_ai', { selection })
```
danvk, line 8: Suggest using `async` / `await` instead of `then`-chaining.

src/views/AIView.tsx
danvk, line 26: We should discuss everyone's preferences around inline styles. I find that all the inline styles in this component make it a bit difficult to follow the logic.

src/views/EditorView.tsx
```diff
@@ -0,0 +1,185 @@
+import { Editor, EditorContent } from '@tiptap/react'
+
+const MenuBar = ({ editor }: { editor: Editor }) => {
+    if (!editor) {
```
danvk, line 4: I don't think `editor` is nullable, so this can be removed.

src/views/ReferencesView.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <p style={{
+                marginTop: "1rem",
+                marginBottom: "1rem",
+                fontStyle: "italic"
+            }}>
+                Click on a reference to add it to the document.
+            </p>
+
+            <ul style={{ borderWidth: "1px" }}>
+                {REFS_DATABASE.map((ref) => (
```
danvk, line 20: Suggest spelling out `reference` since `ref` means something very specific in React.

danvk (commented):
Nice to see this come together!

A few high level comments on the code:

- Did the `CollapsibleBlock` code come from somewhere or is that custom?
- Please run `prettier` on all files.
- We should set up eslint, but I have no preference about whether that happens before or after this PR is merged to `main`.

src/hooks/useDebounce.ts
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+
+export default <T>(value: T, ms: number) => {
```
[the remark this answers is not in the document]
sehyod: I actually implemented both versions, and the second time I figured it made more sense for the hook to debounce a value rather than a callback's result.
As per your suggestion I have now installed [usehooks](https://github.com/juliencrn/usehooks-ts) instead of having our own implementation.

src/CollapsibleBlock/CollapsibleBlock.tsx
sehyod: Unfortunately typing TipTap decorator nodes is not straight forward as they use. I have added a comment and some extra types to prevent having runtime errors

src/views/AIView.tsx
sehyod: We will be using tailwind for style from now on. I only used inline style for the PoC and this will be removed before we merge the PR

src/views/EditorView.tsx
sehyod: Thanks for the catch!

src/views/EditorView.tsx
```diff
@@ -0,0 +1,168 @@
...
+      <button onClick={handleToggleMarkdown}>Toggle Markdown</button>
+    </>
+  );
+};
+
+interface TipTapProps {
+  editor: Editor;
+}
+
+export default ({ editor }: TipTapProps) => {
```
cguedes, line 161: @sehyod avoid this export default. Export a TipTapEditor instead.

sehyod: Nice catch! Thank you

cguedes:
@sergioramos this is ready for review

Issue #18, opened 2023-05-18
TypeScript styles / norms


danvk:
Kicking off a discussion. As a general rule style guidelines should all be automated so that code reviews don't get bogged down by discussion of whitespace preferences.

- Use prettier for code formatting.
  - Everyone should enable "format-on-save"
  - We should check formatting as part of CI
  - Single quotes, two character indent, yes to semicolons, 80 character width (some groups prefer 100)
- Use a recent version of TypeScript with `strict` set
- Use eslint for linting
  - Enable `eslint:recommended`
  - Enable [`@typescript-eslint/recommended'](eslint:recommended', 'plugin:@typescript-eslint/recommended)
  - Enable `plugin:react/recommended` and `plugin:react-hooks/recommended`

Some other suggestions of mine, open to discussion.

General JavaScript:

- Avoid `export default` (named exports encourage consistent use of the same name for the same thing).
- Use a canonical ordering for imports with three (or fewer) groups separated by a blank line, alphabetized within each group (this can be automated with the [`simple-import-sort`](https://github.com/lydell/eslint-plugin-simple-import-sort) eslint plugin):
    ```ts
    import {foo} from 'node-module1';
    import {bar} from 'node-module2';

    import {baz} from './relative-import/a';
    import {quux} from './relative-import/b';

    import './static-asset.css';
    ```
- Use `async` / `await` instead of raw Promises and `.then` whenever possible.

A few TypeScript-specific rules:

- Avoid `any` wherever reasonably possible.
- Prefer `interface` to `type` where possible for simple object types (your type will show up more predictably in error messages and it will improve TypeScript compiler performance for large code bases).
- Prefer `T[]` to `Array<T>` unless `T` is very complex and the latter reads more clearly.
- Avoid non-standard TypeScript language extensions such as `enum` and [parameter properties](https://typescript-eslint.io/rules/parameter-properties/)
- Use synthetic default imports where possible, e.g. `import React from 'react';` instead of `import * as React from 'react';`

@cguedes @sergioramos @sehyod would love feedback and suggestions!

cguedes:
@danvk I agree with everything above. I'm already using almost all of these in my projects. I really like the "format on save" rule.

The only one I would like to change is the character width rule from 80 to *120*.  80 makes sense for prose code writing (ex: typescript) but we will have react and there 120 works best IMO otherwise you can very easily get wrapped to multiple lines.

[...]

PR #24, opened 2023-05-22
chore: add TypeScript styles and norms

cguedes:
Add recommended rules from https://github.com/refstudio/refstudio/issues/18

- 120 char width (prettier)
- Add eslint deps and rules (.eslintrc)
- Fix files using: simple-import-sort
- Manual fixes for any

I've also added the .`/vscode/settings.json` and .`/vscode/extensions.json` with recommended settings. It makes easier to comply with rules after a git clone.


hammer:
Should we use [pre-commit](https://pre-commit.com/) to automate the enforcement of these guidelines?

cguedes:
@hammer this project was already using (from Tauri) [husky](https://typicode.github.io/husky).
I've fixed the commit-hook and now (after `yarn install`) should be working.

I've configured the project to use, _file based_ commit hook (.husky/pre-commit), to run the already configured `lint-staged`.

danvk:
My two cents: I've always preferred pre-push hooks (rather than pre-commit) so that making local commits is instant but your code is always nicely formatted and linted by the time anyone else would have to look at it.

cguedes:
Added pre-push that runs `lint:check`

package.json
```diff
@@ -37,22 +40,22 @@
...
+    "@typescript-eslint/eslint-plugin": "^5.59.6",
+    "@typescript-eslint/parser": "^5.59.6",
     "@vitejs/plugin-react": "^3.0.0",
-    "husky": "^8.0.3",
+    "eslint": "^8.41.0",
+    "eslint-plugin-react": "^7.32.2",
+    "eslint-plugin-react-hooks": "^4.6.0",
+    "eslint-plugin-simple-import-sort": "^10.0.0",
+    "husky": "^8.0.0",
     "lint-staged": "^13.2.0",
```
sehyod, line 51: Do we still need lint-staged if we only setup pre-push hooks?

sehyod (approved):
Looks good to me, thank you for this! I made a comment about `lint-staged`, that I think we can remove, but nothing blocking

cguedes:
This was fixed by #24

Issue #14, opened 2023-05-16
Tooling for PDF Ingestion


gjreda:
The backend side of this project is comprised of two components (for now):
1. PDF Ingestion
2. AI Interactions

This issue is meant to capture work and discussion related to the backend side of PDF Ingestion.

### Initial Exploration of Python PDF parsing libraries
Candidates:
- [PyPDF](https://github.com/py-pdf/pypdf) - a pure Python PDF parser (which I think used to be called PyPDF2)
- [pdfplumber](https://github.com/jsvine/pdfplumber) - which builds on [pdfminer.six](https://github.com/pdfminer/pdfminer.six). This library is used in AI2's `mmda` and [vila](https://github.com/allenai/vila).
- [PyMuPDF](https://github.com/pymupdf/PyMuPDF), which is what Haystack's PDF ingestion builds on
- [mmda](https://github.com/allenai/mmda/) - a work-in-progress from AI2, and uses `pdfplumber` and [pdf2image](https://github.com/Belval/pdf2image) under the hood
- [grobid](https://github.com/kermitt2/grobid) - used by [s2orc-doc2json](https://github.com/allenai/s2orc-doc2json). Written in Java, which would mean running another tauri-sidecar.
- [s2orc-doc2json](https://github.com/allenai/s2orc-doc2json) - Python tool that builds on grobid and is used by AI2's Semantic Scholar. Notably, this would require also having a grobid process running on a refstudio user's machine (and making sure Java is installed).

### pypdf
I used this for prototyping the pdf chatbot in [scratch-pdf-bot](https://github.com/gjreda/scratch-pdf-bot). It did a fine job at extracting text, but I didn't delve too deep beyond that. Seems fairly similar to pdfplumber, with the one notable difference that pdfplumber has additional functionality for table extraction.

### mmda
As noted in its readme, this library seems still in its early development. Installing from `pypi` via `poetry add`, I ran into `ModuleNotFoundError` exceptions trying to run the code snippet provided in the readme (missing dependencies for tokenizers, layoutparser, vila) and also noted an empty `setup.py`.

I think this project is worth continuing to watch, but don't think it's a good fit for refstudio right now.

----
Given its functionality and adoption in other projects, I'm going to explore `pdfplumber` further.

hammer:
Not urgent but in case we decide to move this into Rust https://github.com/pdf-rs/pdf looks relevant

gjreda:
[...]

PR #29, opened 2023-05-23
feat: run PDF ingestion on document upload

gjreda:
- https://github.com/refstudio/refstudio/issues/14

Including the new dependencies in the binary was rather finicky, so please pull down this branch and try to test it. I want to make sure it works on more machines than mine.

To do so:
```bash
$ poetry install
$ yarn python
$ yarn tauri dev
```

Note that this might take a minute or two to run once a document is uploaded. You can check the `.lancedb` directory for your document to make sure the process has completed successfully.
```bash
$ ls -la ~/.ref-studio/project-x/.lancedb
total 0
drwxr-xr-x  4 greg  staff  128 May 23 14:53 .
drwxr-xr-x  9 greg  staff  288 May 23 14:52 ..
drwxr-xr-x  5 greg  staff  160 May 23 14:52 A Few Useful Things to Know about Machine Learning.tei.lance
drwxr-xr-x  5 greg  staff  160 May 23 14:53 Machine Learning at Scale.tei.lance
```

<hr>

When a document is uploaded, run the PDF ingestion pipeline so that we can perform Q&A over the documents.

Right now, the pipeline stores everything in the project working directory, but I suspect we'll want to change that later (there's little reason for the user to see this data):
1. Call the HF grobid server, writing the output back to a `grobid` directory within the project working directory
2. Convert the XML to JSON for easier parsing down the road
3. Use `sentence-transformers` to generate sentence embeddings for dense retrieval during AI interactions
4. Store embeddings via `lancedb` -- we will query this data in AI interactions

Some things to note:
1. Since this pipeline will not be instantaneous, we'll ideally get the user to upload all their PDFs at once (or point to a directory).
2. Both `sentence-transformers` and `lancedb` have some very hefty dependencies (torch, pyarrow). These increase the size of the binary significantly and slow down calls to it. This is something we'll want to optimize once we have the full end-to-end flow in place. We might even want to remove the dependency and use OpenAI to generate the embeddings.
3. The `chunk` strategy for documents is terribly naive right now and is worth improving later.
4. I'd like to get the end-to-end ingestion and Q&A flow in place first, so this does pull additional document metadata into lancedb (yet ... it's all in the json though).

image


src-tauri/src/main.rs
```diff
@@ -4,16 +4,30 @@
 // Learn more about Tauri commands at https://tauri.app/v1/guides/features/command
```
gjreda, line 4: Hmm, I didn't change anything with this file 🤔

gjreda:
Think the diff on this PR should shrink significantly once @cguedes merges his here: https://github.com/refstudio/refstudio/pull/25. This branch is based off of that one.

src/views/FoldersView.tsx
```diff
@@ -33,6 +33,10 @@ export function FoldersView({ onClick }: { onClick?: (fileEntry: FileEntry) => v
       readAllProjectFiles().then(setFiles);
       console.log(files);
     });
+    runPDFIngestion().then(() => {
+      console.log('PDFs ingested with success');
+      readAllProjectFiles().then(setFiles);
+    });
```
gjreda, line 39: @cguedes @sehyod Advice or thoughts on how to do this piece are much appreciated (my JS/TS/React knowledge is minimal). I'll look into a way of communicating ingestion progress, but for now maybe we can just show a spinner or something off to the side until the sidecar command returns.

hammer:
It might be worth making this PR inclusive of adding the newly parsed PDF document to the list of references in the references component.

hammer:
Also as a reminder please reference the associated issue in the PR description!

hammer:
Copying over some comments from Slack to have them on GitHub:

> I'm less concerned with install size and more concerned with resident memory size, and especially memory leaks, cf. this terrifying thread https://github.com/tauri-apps/tauri/issues/4026
> 
> But I hear you, if we end up shipping multiple models to the client we're going to be really bloating the download
> 
> BTW the several seconds of lag for each call is possibly because the model has to be loaded into memory for each call since we're calling the sidecar as a command-line binary not as a service. I may be wrong as I haven't fully grokked the Tauri architecture. It will be interesting to see if the lag goes down with WebSockets. Longer term we can also explore various inference optimization approaches such as TVM, Neural Magic, etc.

src/views/FoldersView.tsx
sehyod: About the way to communicate progress, I think this is related to #33, we need to define the way we want to pass information between components, in this case the loading status/progress.

About how to do this piece, I guess it depends on when we want to run the ingestion process. For now, if we want to run it every time files are uploaded, that is the right place to do it. However, you need to call `runPDFIngestion` once files have been uploaded otherwise you will run into race conditions with the sidecar being called before the app directory receives the new files. I'm not a big fan of the `.then` chaining syntax and prefer the async/await, so you could write something like this:
```typescript
const handleChange = async (files: FileList) => {
    await uploadFiles(files);
    console.log('File uploaded with success');
    console.log(files);
    let updatedFiles = await readAllProjectFiles();
    setFiles(updatedFiles);

    await runPDFIngestion()
    console.log('PDFs ingested with success');
   updatedFiles = await readAllProjectFiles();
    setFiles(updatedFiles);
  };
```

cguedes: I would provide `runPDFIngestion` with both the current array of file paths uploaded, and also the base directory for the uploads. Both absolute paths to the filesystem. 

This would allow the ingestion to either read new files uploaded or the full list of files in the upload folder.

```js
// In filesystem.ts
export async function getUploadsDir() {
  return join(await getBaseDir(), UPLOADS_DIR);
}

// In FoldersView.tsx
const handleChange = async (files: FileList) => {
    await uploadFiles(files);
    console.log('File uploaded with success');
    console.log(files);
    let updatedFiles = await readAllProjectFiles();
    setFiles(updatedFiles);

    const uploadedFilesPath = updatedFiles.map(f => f.path)
    const uploadsDir = await getUploadsDir()
    await runPDFIngestion(uploadedFilesPath, uploadsDir)
    console.log('PDFs ingested with success');
    updatedFiles = await readAllProjectFiles();
    setFiles(updatedFiles);
  };
```

gjreda:
> It might be worth making this PR inclusive of adding the newly parsed PDF document to the list of references in the references component.

Something we can discuss more tomorrow, but the ingestion process writes json files to the filesystem, so one option is for the sidecar to return the filepath for the json file, which the client side can then read from the filesystem.

gjreda:
Pushed an update so that this process now returns the uploaded reference metadata json once it has completed. I think it makes sense to separate this PR (backend ingestion) from its frontend presentation (I've created https://github.com/refstudio/refstudio/issues/49).

```bash
$ ./src-tauri/bin/python/main-aarch64-apple-darwin/main ingest --pdf_directory="/Users/greg/Library/Application Support/com.tauri.dev/project-x/uploads" | jq
{
  "project_name": "project-x",
  "references": [
    {
      "title": "Machine Learning at Scale",
      "authors": [
        {
          "full_name": "Sergei Izrailev",
          "given_name": "Sergei",
          "surname": "Izrailev",
          "email": "sizrailev@collective.com"
        },
        {
          "full_name": "Jeremy M Stanley",
          "given_name": "Jeremy",
          "surname": "Stanley",
          "email": "jstanley@collective.com"
        }
      ]
    },
    {
      "title": "A Few Useful Things to Know about Machine Learning",
      "authors": [
        {
          "full_name": "Pedro Domingos",
          "given_name": "Pedro",
          "surname": "Domingos",
          "email": "pedrod@cs.washington.edu"
        }
      ]
    },
    {
      "title": "Hidden Technical Debt in Machine Learning Systems",
      "authors": [
        {
          "full_name": "D Sculley",
          "given_name": "D",
          "surname": "Sculley",
          "email": "dsculley@google.com"
        },
        {
          "full_name": "Gary Holt",
          "given_name": "Gary",
          "surname": "Holt",
          "email": "gholt@google.com"
        },
        {
          "full_name": "Daniel Golovin",
          "given_name": "Daniel",
          "surname": "Golovin",
          "email": null
        },
        {
          "full_name": "Eugene Davydov",
          "given_name": "Eugene",
          "surname": "Davydov",
          "email": "edavydov@google.com"
        },
        {
          "full_name": "Todd Phillips",
          "given_name": "Todd",
          "surname": "Phillips",
          "email": "toddphillips@google.com"
        },
        {
          "full_name": "Dietmar Ebner",
          "given_name": "Dietmar",
          "surname": "Ebner",
          "email": "ebner@google.com"
        },
        {
          "full_name": "Vinay Chaudhary",
          "given_name": "Vinay",
          "surname": "Chaudhary",
          "email": "vchaudhary@google.com"
        },
        {
          "full_name": "Michael Young",
          "given_name": "Michael",
          "surname": "Young",
          "email": "mwyoung@google.com"
        },
        {
          "full_name": "Jean-Franc ¸ois Crespo",
          "given_name": "Jean-Franc ¸ois",
          "surname": "Crespo",
          "email": "jfcrespo@google.com"
        },
        {
          "full_name": "Dan Dennison",
          "given_name": "Dan",
          "surname": "Dennison",
          "email": "dennison@google.com"
        }
      ]
    }
  ]
}
```

gjreda:
Ok, this should be good to merge now.

Some related action items:
1. It is slow due to `PyInstaller --onefile`. But, if I get rid of this flag, we get [the errors I mentioned in Slack](https://ref-studio.slack.com/archives/C05525DV9HN/p1685124347714859). I've filed issues #56 and #48 to investigate further.
2. Issue #50 to separate embedding creation from PDF ingestion. Right now it is all one linear pipeline, but it makes sense to break up.
3. Issue #49 to populate the ReferenceView with the response from PDF ingestion. I'll leave this one to either @cguedes or @sehyod.

image

scripts/reset_uploads.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+PROJECT_DIR="/Users/greg/Library/Application Support/com.tauri.dev/project-x"
```
cguedes, line 3: @gjreda was this a temp reset script or should we have one and make it user-agnostic?

cguedes (dismissed):

cguedes (approved):

PR #30, opened 2023-05-24
Adds Redux store

cguedes:
This PR adds [redux](https://react-redux.js.org/) using [redux toolkit](https://redux-toolkit.js.org/) to simplify the application data flow and closes issue #33 

**Slices**

With this change, two slices were created: 

* `selection` - store for the selected text in the editor
* `openedFiles` - store for the opened files, and the selected file.

**UI Component Changes**

With the addition of the redux store, the following components were changed/added:
* `TipTapEditor` - to dispatch selection changes (debounced)
* `AIView` - use the store instead of props
* `FoldersView` - use the store instead of props
* `CenterPaneView` - rewrite to display different types of center views depending on the selected file
* `OpenedFilesView` - added to display and manage the selected files. It allow file selection and close

**Current UI**

image

### DIFF
--- a/src/TipTapEditor/TipTapEditor.tsx
+++ b/src/TipTapEditor/TipTapEditor.tsx
@@ -9,7 +9,7 @@ import { setTextSelection } from '../features/selection/selectionSlice';
 import { EditorProps } from '../types/EditorProps';
 import { MenuBar } from './MenuBar';
 import { ReferenceNode } from './ReferenceBlock/ReferenceNode';
-import { EDITOR_EXTENSIONS, INITIAL_CONTENT } from './TipTapEditorConfigs';
+import { EDITOR_EXTENSIONS } from './TipTapEditorConfigs';
 
 export function TipTapEditor({ editorRef, editorContent }: EditorProps) {
   const [selection, setSelection] = useState('');
@@ -19,7 +19,7 @@ export function TipTapEditor({ editorRef, editorContent }: EditorProps) {
     setEditor(
       new Editor({
         extensions: EDITOR_EXTENSIONS,
-        content: editorContent ?? INITIAL_CONTENT,
+        content: editorContent,
         onSelectionUpdate({ editor }) {
           const { from, to } = editor.view.state.selection;
           const text = editor.view.state.doc.textBetween(from, to);
--- a/src/TipTapEditor/TipTapEditor.tsx
+++ b/src/TipTapEditor/TipTapEditor.tsx
@@ -2,24 +2,39 @@ import './TipTapEditor.css';
 
 import { Editor, EditorContent } from '@tiptap/react';
 import { useEffect, useState } from 'react';
+import { useDispatch } from 'react-redux';
+import { useDebounce } from 'usehooks-ts';
 
+import { setTextSelection } from '../features/selection/selectionSlice';
 import { EditorProps } from '../types/EditorProps';
 import { MenuBar } from './MenuBar';
 import { ReferenceNode } from './ReferenceBlock/ReferenceNode';
 import { EDITOR_EXTENSIONS, INITIAL_CONTENT } from './TipTapEditorConfigs';
 
-export function TipTapEditor({ editorRef, editorContent, onSelectionChange }: EditorProps) {
+export function TipTapEditor({ editorRef, editorContent }: EditorProps) {
+  const [selection, setSelection] = useState('');
+
   const [editor, setEditor] = useState<Editor | null>(null);
   useEffect(() => {
-    setEditor(new Editor({
-      extensions: EDITOR_EXTENSIONS,
-      content: editorContent ?? INITIAL_CONTENT,
-      onSelectionUpdate({ editor }) {
-        const { from, to } = editor.view.state.selection;
-        onSelectionChange(editor.view.state.doc.textBetween(from, to));
-      },
-    }));
-  }, [editorContent, onSelectionChange]);
+    setEditor(
+      new Editor({
+        extensions: EDITOR_EXTENSIONS,
+        content: editorContent ?? INITIAL_CONTENT,
+        onSelectionUpdate({ editor }) {
+          const { from, to } = editor.view.state.selection;
+          const text = editor.view.state.doc.textBetween(from, to);
+          setSelection(text);
+        },
+      }),
+    );
+  }, [editorContent]);
+
+  // Dispatch (debounced) selection changed
+  const dispatch = useDispatch();
+  const debouncedSelection = useDebounce(selection, 200);
+  useEffect(() => {
+    dispatch(setTextSelection(debouncedSelection));
+  }, [dispatch, debouncedSelection]);
 
   useEffect(() => {
     if (!editor) return;
--- a/src/features/openedFiles/openedFilesSlice.ts
+++ b/src/features/openedFiles/openedFilesSlice.ts
@@ -1,0 +1,59 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { FileEntry } from '@tauri-apps/api/fs';
+
+type FileEntryType = 'TipTap' | 'PDF' | 'Unknown';
+
+export interface FileEntryInfo {
+  entry: FileEntry;
+  type: FileEntryType;
+  isDirectory: boolean;
+}
+
+export interface OpenedFilesState {
+  selectedFile?: FileEntryInfo;
+  files: FileEntryInfo[];
+}
+
+const initialState: OpenedFilesState = {
+  selectedFile: undefined,
+  files: [],
+};
+
+export const openedFilesSlice = createSlice({
+  name: 'openedFiles',
+  initialState,
+
+  reducers: {
+    openFile: (state, action: PayloadAction<FileEntryInfo>) => {
+      const exists = state.files.find(sameFileAs(action.payload));
+      if (!exists) {
+        state.files.push(action.payload);
+      }
+      state.selectedFile = state.files.find(sameFileAs(action.payload));
+    },
+
+    selectFile: (state, action: PayloadAction<FileEntryInfo>) => {
+      state.selectedFile = state.files.find(sameFileAs(action.payload));
+    },
+
+    closeFile: (state, action: PayloadAction<FileEntryInfo>) => {
+      state.files = state.files.filter((e) => e.entry.path !== action.payload.entry.path);
+      if (sameFile(state.selectedFile, action.payload)) {
+        state.selectedFile = state.files[0];
+      }
+    },
+  },
+});
+
+export const { openFile, selectFile, closeFile } = openedFilesSlice.actions;
+
+export default openedFilesSlice.reducer;
+
+function sameFile(a?: FileEntryInfo, b?: FileEntryInfo) {
+  if (!a || !b) return false;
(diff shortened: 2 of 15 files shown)


src/views/AIView.tsx
```diff
@@ -1,7 +1,10 @@
 import { Command } from '@tauri-apps/api/shell';
 import { useEffect, useState } from 'react';
 
-export function AIView({ selection }: { selection: string | null }) {
+import { useAppSelector } from '../redux/hooks';
+
+export function AIView({ selection: x }: { selection: string | null }) {
```
sehyod: You should remove `selection` from the props, we aren't using it anymore.

src-tauri/src/main.rs
```diff
@@ -16,13 +16,14 @@ fn interact_with_ai(selection: &str) -> String {
...
 use tauri::Manager;
+use std::env;
 
 fn main() {
     tauri::Builder::default()
         .setup(|app| {
             #[cfg(debug_assertions)] // only include this code on debug builds
             {
-                let dev_tools_visible = false;
+                let dev_tools_visible = env::var("DEV_TOOLS").is_ok();
```
sehyod, line 26: Where is this variable defined? Is it in a .env file? And if so, could we add a .env.template file so anyone cloning the repo knows what are the environment variables we are using?

src/views/CenterPaneView.tsx
```diff
@@ -21,18 +21,25 @@ export function CenterPaneView({ file, ...props }: CenterPaneViewProps) {
         const textContent = new TextDecoder('utf-8').decode(content);
         setContent(textContent);
         setLoading(false);
-      })()
+      })();
```
sehyod, line 24: Unrelated to this PR, but it's weird we have been able to push the code without the semi-colon. Are the pre-push hooks working properly?

sehyod (approved):

src/views/CenterPaneView.tsx
cguedes: Done.

src-tauri/src/main.rs
cguedes: Yes, that's an env file. We can also just add a new script to package json like `yarn tauri:dev:tools` or `yarn tauri:dev:debug` what do you think?

src/views/AIView.tsx
cguedes: That was a typo. Good catch.

src/views/CenterPaneView.tsx
```diff
@@ -1,39 +1,46 @@
...
-  if (loading)
-    return (
-      <div>
-        <strong>Loading...</strong>
-      </div>
-    );
+    setLoading(true);
+    (async () => {
+      // Note: Somehow I need need to await a bit for the TipTap to work!!
+      await new Promise((resolve) => setTimeout(resolve, 100));
```
sehyod, line 31: This is something we should probably investigate further. Maybe you could create an issue about it.

sehyod (dismissed):

src/views/CenterPaneView.tsx
cguedes: You are right. I will create an issue for this.

hammer:
Nice work! In general I prefer to keep PRs restricted to a single coherent set of changes, so in the future please break out the startup scripts, TailwindCSS, and git hook, and changes separately if it's not too much trouble.

cguedes:
> Nice work! In general I prefer to keep PRs restricted to a single coherent set of changes, so in the future please break out the startup scripts, TailwindCSS, and git hook changes separately if it's not too much trouble.

You are right. I will do my best to scope these changes in separate PRs.

src/views/OpenedFilesView.tsx
```diff
@@ -0,0 +1,57 @@
...
+      )}
+    >
+      {openedFiles.files.map((fileInfo) => (
+        <span
+          onClick={(evt) => {
+            evt.stopPropagation();
+            dispatch(selectFile(fileInfo));
+          }}
+          className={cx(
+            'flex flex-wrap items-center justify-between',
```
danvk, line 27: With the caveat that I haven't used TailwindCSS before so it's hard for me to judge whether this is normal… the logic of the component really feels like it's getting drowned out by all these class names. Would it make sense to factor out a meaningfully-named component that applies most of these styles? https://tailwindcss.com/docs/reusing-styles#extracting-components-and-partials

danvk (commented):
It was interesting to learn about Redux Toolkit in the process of reviewing this PR! I left a lot of comments, but at a high level:

- I do think it's a good idea to have some sort of state management tool for truly global application state.
- That tool may or may not be RKT, as per our discussion of Recoil earlier today.
- Even with a state management library, we should try to:
  1. Use built-in state management (`useState`) for state that's local to a component.
  2. Keep low-level components presentational (store agnostic) for better encapsulation / separation of concerns.
  3. Avoid multiple sources of truth (i.e. state in the global store and in local state).

src/TipTapEditor/TipTapEditor.tsx
```diff
@@ -2,24 +2,39 @@ import './TipTapEditor.css';
...
+import { useDebounce } from 'usehooks-ts';
 
+import { setTextSelection } from '../features/selection/selectionSlice';
 import { EditorProps } from '../types/EditorProps';
 import { MenuBar } from './MenuBar';
 import { ReferenceNode } from './ReferenceBlock/ReferenceNode';
-import { EDITOR_EXTENSIONS, INITIAL_CONTENT } from './TipTapEditorConfigs';
+import { EDITOR_EXTENSIONS } from './TipTapEditorConfigs';
+
+export function TipTapEditor({ editorRef, editorContent }: EditorProps) {
```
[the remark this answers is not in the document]
cguedes: I agree. We did this to suppport multiple editor implementation(i.e. Lexical) and we wanted to keep a common interface, but we can drop that for now.

src/TipTapEditor/TipTapEditor.tsx
```diff
@@ -2,24 +2,39 @@ import './TipTapEditor.css';
...
-    setEditor(new Editor({
-      extensions: EDITOR_EXTENSIONS,
-      content: editorContent ?? INITIAL_CONTENT,
-      onSelectionUpdate({ editor }) {
-        const { from, to } = editor.view.state.selection;
-        onSelectionChange(editor.view.state.doc.textBetween(from, to));
-      },
-    }));
-  }, [editorContent, onSelectionChange]);
+    setEditor(
```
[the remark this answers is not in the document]
cguedes: We need to refactor this code (interaction with TipTap), but unfortunately their API to set/reset content isn't straightforward. CC @sehyod

src/TipTapEditor/TipTapEditor.tsx
```diff
@@ -2,24 +2,39 @@ import './TipTapEditor.css';
...
+  }, [editorContent]);
+
+  // Dispatch (debounced) selection changed
+  const dispatch = useDispatch();
+  const debouncedSelection = useDebounce(selection, 200);
+  useEffect(() => {
+    dispatch(setTextSelection(debouncedSelection));
+  }, [dispatch, debouncedSelection]);
 
   useEffect(() => {
```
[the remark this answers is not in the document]
cguedes: We moved that code already to the receiving party and it's now more simple to understand.
Also, this is only "sample interaction code" between the editor and the AI Panel.

src/features/openedFiles/openedFilesSlice.ts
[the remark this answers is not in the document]
cguedes: I wasn't sure we would have more files for each feature. But agree we should start with a simpler tree.

src/features/openedFiles/openedFilesSlice.ts
```diff
@@ -0,0 +1,59 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { FileEntry } from '@tauri-apps/api/fs';
+
+type FileEntryType = 'TipTap' | 'PDF' | 'Unknown';
+
+export interface FileEntryInfo {
+  entry: FileEntry;
+  type: FileEntryType;
```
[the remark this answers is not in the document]
cguedes: Exactly. I wasn't sure if this was the correct approach. 
I wanted to create a wrapper over the file entry with additional properties (ex: type) but wasn't sure how "complex" the TS should be.

src/views/FoldersView.tsx
```diff
@@ -41,22 +41,14 @@ export function FoldersView({ onClick }: { onClick?: (fileEntry: FileEntry) => v
...
-interface FileTreeBaseProps {
-  root?: boolean;
-  files: FileEntry[];
-  file?: FileEntry;
-  selectedFile?: FileEntry;
-  onClick: (file: FileEntry) => void;
-}
-
-const FileTree = ({ files, root, ...props }: FileTreeBaseProps) => {
+function FileTree({ files, root = false }: { files: FileEntry[]; root?: boolean }) {
```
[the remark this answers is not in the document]
cguedes: Sure, I think that we would build more code but in the end it pays off as we can test the UI component in isolation from the store, and we can also replace UI elements (ex: TreeNode) by libraries.

> Keep low-level components presentational (store agnostic) for better encapsulation / separation of concerns.

src/views/FoldersView.tsx
```diff
@@ -65,17 +57,34 @@ const FileTree = ({ files, root, ...props }: FileTreeBaseProps) => {
...
+  const dispatch = useAppDispatch();
+  const selectedFile = useAppSelector((state) => state.openedFiles.selectedFile);
 
-const FileTreeNode = ({ file, onClick, selectedFile }: FileTreeBaseProps) => {
   if (!file) return null;
   const isFolder = Array.isArray(file.children);
 
+  function handleOnClick(file: FileEntry): void {
+    dispatch(
+      openFile({
```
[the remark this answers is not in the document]
cguedes: I see, that's in line with the change of refactoring the component to be more isolated.

src/views/OpenedFilesView.tsx
```diff
@@ -0,0 +1,57 @@
...
+            'cursor-pointer select-none whitespace-nowrap',
+            'px-2 py-1', //
```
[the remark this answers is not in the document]
cguedes: This is for legibility. That would make prettier wrap the array in multiple lines, even with fewer statements.
Also, the idea is to break (using cx) a long list of rules into simpler rules/lines. This also makes diff simpler.

src/views/OpenedFilesView.tsx
cguedes: Sure. The focus of this PR was adding the state management (Redux) so I didn't added much ❤️ into that code. For the IDE issue I will fix this.

src/TipTapEditor/TipTapEditor.tsx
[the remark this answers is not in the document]
sehyod: Yes, unfortunately, using `setText` leads to unwanted behaviours, especially regarding the history, because there is no easy way to clear the history (cf https://github.com/ueberdosis/tiptap/issues/491).
I think at some point we will have in-memory editors anyway, corresponding to open tabs, so the create operation would only be called once per file.

cguedes:
Closing this as we decided not to use Redux for this project. See #33 for more details.

Issue #28, opened 2023-05-23
View PDF files inline


hammer:
@cguedes mentioned he's interested in exploring our options for adding PDF viewing.

As mentioned in last week's sync meeting, https://github.com/allenai/pdf-component-library is a library from AI2 built on https://github.com/wojtekmaj/react-pdf that may be useful for us. I believe this library is used in their [Semantic Reader](https://www.semanticscholar.org/product/semantic-reader) project.

cguedes:
This was closed by #31

Issue #1, opened 2023-05-03
Design high-level layout of writing, references, and AI interactions components


hammer:
For familiarity we need to decide if this should feel like a word processor (WP) or an IDE.

1. WP: Microsoft Word, Google Docs
2. IDE: VS Code, RStudio, JetBrains

For the WP, we'd have a persistent left pane with references and right pane with AI interactions.
For the IDE, we'd have collapsible and relocatable windows for writing, references, and AI interactions.

hammer:
Pulling some comments out of PRs to track:

- We can resize components now. How easy is it to collapse them entirely, or re-arrange them?
- What is the difference between "Uploads" and "References"? Do we want them to be separate?
- Vertical scroll within a component not working yet

gjreda:
> What is the difference between "Uploads" and "References"? Do we want them to be separate?

No strong opinion, but I could imagine an "Upload" is the raw document (an unstructured PDF), whereas a Reference is the structured JSON we create from the grobid response

<summary>Example citations list from Grobid</summary>
<details>

```
[
  {
    "authors": [
      {
        "full_name": "References",
        "surname": "References"
      }
    ],
    "index": 0,
    "id": "b0"
  },
  {
    "authors": [
      {
        "full_name": "O Chapelle",
        "given_name": "O",
        "surname": "Chapelle"
      },
      {
        "full_name": "E Manavoglu",
        "given_name": "E",
        "surname": "Manavoglu"
      },
      {
        "full_name": "R Rosales",
        "given_name": "R",
        "surname": "Rosales"
      }
    ],
    "index": 1,
    "id": "b1",
    "date": "2014",
    "title": "A simple and scalable response prediction for display advertising",
    "journal": "ACM Trans. Intell. Syst. Technol",
    "note": "To appear"
  },
  {
    "authors": [
      {
        "full_name": "K Lee",
        "given_name": "K",
        "surname": "Lee"
      },
      {
        "full_name": "B Orten",
        "given_name": "B",
        "surname": "Orten"
      },
      {
        "full_name": "A Dasdan",
        "given_name": "A",
        "surname": "Dasdan"
      },
      {
        "full_name": "W Li",
        "given_name": "W",
        "surname": "Li"
      }
    ],
    "index": 2,
    "id": "b2",
    "date": "2012",
    "title": "Estimating conversion rate in display advertising from past performance data",
[...]

PR #31, opened 2023-05-24
feat: view PDF files inline

sehyod:
This PR adds a PDF viewer component, as mentioned in #28 

To make this PR, I explored https://github.com/allenai/pdf-component-library and the library it's built upon: https://github.com/wojtekmaj/react-pdf.

With `pdf-component-library`, I wasn't able to properly render a PDF: the best I was able to do was displaying the raw text of the pdf, but all styling would be gone. Moreover, there is no documentation for the library and it doesn't seem maintained anymore (latest commit was in December).

On the other hand, `react-pdf`'s doc looks more exhaustive and the library is still maintained (latest commit is from yesterday). With this library, I was able to properly render PDF files:
image

For these reasons, I have decided to go with `react-pdf`, although `pdf-component-library` offers more possibilities, like thumbnails and annotations. I think it's fine to go with `react-pdf` for the MVP.


hammer:
@sehyod very cool! This looks to share some commits with https://github.com/refstudio/refstudio/pull/29. In the future, could we break the PRs up so that the drag and drop zone is separate from the PDF viewer? That way PRs are more isolated and it's easier to collaborate and stack PRs from separate developers.

Could you say more about the implementation strategy here? Any issues with library selection or concerns about resource utilization or leakage? I got it running and it looks nice. I was a bit nervous about having this feature now as it's not high priority but it looks simple lightweight and useful so let's keep it!

cguedes:
My first notes about the PR are that:

- by default the PDF view don't resize the loaded document
- links inside the document open in the full Tauri view instead of the PDF (or externally in the system browser)

src/App.tsx
```diff
@@ -1,22 +1,30 @@
...
 import { AIView } from './views/AIView';
 import { CenterPaneView } from './views/CenterPaneView';
 import { FoldersView } from './views/FoldersView';
 import { ReferencesView } from './views/ReferencesView';
 
+pdfjs.GlobalWorkerOptions.workerSrc = new URL(
+  'pdfjs-dist/build/pdf.worker.min.js',
+  import.meta.url,
+).toString();
+
```
cguedes, line 20: I think that this should be moved outside the `App.tsx` file. Can we move it to the `PdfViewer`?

src/App.tsx
```diff
@@ -25,15 +33,19 @@ function App() {
     setSelectedFile(file);
   }, []);
 
+  const handleCenterPanelResize = () => {
+    pdfViewerRef.current?.updateWidth();
```
cguedes, line 37: I feel that we can inline this

src/PdfViewer/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+import { readFile } from "../filesystem";
+import { PdfViewerAPI } from '../types/PdfViewerAPI';
+
+export interface PdfViewerProps {
+  file: FileEntry;
+  pdfViewerRef: React.MutableRefObject<PdfViewerAPI | null>;
+}
+
+export const PdfViewer = ({ file, pdfViewerRef }: PdfViewerProps) => {
```
cguedes, line 16: I think we should use a function declaration for react components instead of this syntax.

src/PdfViewer/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+import 'react-pdf/dist/esm/Page/TextLayer.css';
+
+import { FileEntry } from "@tauri-apps/api/fs";
+import { useCallback, useEffect, useRef, useState } from "react";
+import { Document, Page } from "react-pdf";
+
+import { readFile } from "../filesystem";
+import { PdfViewerAPI } from '../types/PdfViewerAPI';
+
+export interface PdfViewerProps {
```
cguedes, line 11: You don't need to export this interface.

src/PdfViewer/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [fileContent, setFileContent] = useState<{ data: Uint8Array } | null>(null);
+  const [isFileLoading, setIsFileLoading] = useState(true);
+  const [error, setError] = useState<string | null>(null);
+
+  const containerRef = useRef<HTMLDivElement>(null);
+  const [pdfViewerWidth, setPdfViewerWidth] = useState<number>();
+
+  const updateTimeoutRef = useRef<NodeJS.Timeout>();
+
+  const updateWidth = useCallback(() => {
```
cguedes, line 29: Have you tried to use https://usehooks-ts.com/react-hook/use-timeout?

hammer:
> links inside the document

I think the right behavior would be to open a new tab within the application for the linked PDF.

src/PdfViewer/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+    clearTimeout(updateTimeoutRef.current);
+    updateTimeoutRef.current = setTimeout(() => {
+      setPdfViewerWidth(containerRef.current?.getBoundingClientRect().width);
+    }, 200);
+  }, [updateTimeoutRef]);
+
+  // Update viewer's width on window resize
+  useEffect(() => {
```
cguedes, line 40: Can't we use the https://usehooks-ts.com/react-hook/use-element-size instead of this effect?

src/PdfViewer/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+        } else {
+          console.error(err);
+          setError('Unknown error');
+        }
+      }
+    })()
+  }, [file]);
+
+  const onDocumentLoadSuccess = ({ numPages }: { numPages: number }) => {
+    setNumPages(numPages);
```
cguedes, line 77: I think we can inline this

src/PdfViewer/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+  return (
+    <div className="pdf-viewer flex flex-col h-full" ref={containerRef}>
+      <Document
+        file={fileContent}
+        onLoadSuccess={onDocumentLoadSuccess}
+        className="flex-1 overflow-scroll"
+      >
+        {numPages && Array.from(new Array(numPages), (_, index) => (
+          <Page key={`page_${index + 1}`} pageNumber={index + 1} width={pdfViewerWidth} />
+        ))}
```
cguedes, line 93: Why do we need this part?

cguedes (changes requested):
Added notes as comment before.

cguedes (changes requested):
Added notes as comment before.

cguedes:
> > links inside the document
> 
> I think the right behavior would be to open a new tab within the application for the linked PDF.

Like with an `<iframe ...`? @hammer

hammer:
@cguedes no, as in the comments at https://github.com/refstudio/refstudio/issues/1#issuecomment-1562803431:
> We can then open documents for writing or PDFs for reading in the center in a tabbed configuration

Screenshot from VS Code of 2 tabs:
Screenshot 2023-05-26 at 12 33 58 PM

cguedes:
Sure, in a new tab (maybe in split view).
I mean using an `IFrameTab` component (implemented with an `iframe`) as opposed to use an `EditorTab` or `PdfViewTab` because the type of content to open is an URL.

hammer:
Ah I see yeah not sure we want to grow a full web browser! Would it be better to send the user to their OS browser?

cguedes:
@hammer yes, I like that option the most. @sehyod please check how we can have that.

src/App.tsx
sehyod: Yes, sorry, it was there because of my initial experiments with the library and I forgot to move it afterwards. Thanks for the catch!

src/App.tsx
sehyod: I usually prefer the bracket syntax for void functions because the inline version would be equivalent to
```typescript
const handleCenterPanelResize = () => {
    return pdfViewerRef.current?.updateWidth();
}
```
which is semantically different. Here it wouldn't change anything because `updateWidth` is a void method so it returns undefined. What are you thoughts on that? Should we inline it anyway?

src/PdfViewer/index.tsx
sehyod: This comes from the documentation: https://github.com/wojtekmaj/react-pdf/wiki/Recipes#display-all-pages

Basically, `Document` is just a context provider and displays nothing. Then you have to add the `Page`s. You can either have a the current page as state in your component and only display this page, along with buttons to change the current page, or you display all the pages at once (which is what we want here). Unfortunately, javascript has no clean way to create an array [1, ..., numPages], which is why we have to do the `Array.from(new Array(numPages), (_, index) => (...` "hack".

sehyod:
Thank you for your comments! I have tackled comments about the code. For the issue with links, I will investigate it but the library we are using simply creates hyperlinks and I'm not sure we can customize this.

cguedes (changes requested):

cguedes (approved):

Issue #26, opened 2023-05-23
Python styles / norms


hammer:
Same discussion as https://github.com/refstudio/refstudio/issues/18 but for Python.

gjreda:
@hammer Never tried it, but willing to. I'm into the idea of it.

hammer:
We've grappled with it quite a bit for https://github.com/pystatgen/sgkit, but most of the issues come in from the numerical libraries. In general I think types are critical for a scalable codebase but if you haven't used it in production we can hold off for now.

gjreda:
Bumped `line-length = 120` in PR #29. ruff's default of 88 felt restrictive and we have prettier's line length set to 120. Felt like a fine starting point.

cguedes:
@gjreda I'm closing this issue as I think it's already implemented and validated by ruff. Please reopen otherwise.

danvk:
One Python style thing I wanted to flag: I recently learned that `typing.List[Type]` and friends have been [deprecated since Python 3.9][1] in favor of `list[Type]`:

- `List[T]` -> `list[T]`
- `Dict[K, V]` -> `dict[K, V]`
- `Optional[T]` -> `T | None`

Since we require Python 3.9+ we should just use the newer syntax.

[1]: https://stackoverflow.com/questions/72101566/should-i-use-a-capital-l-list-for-type-hinting-in-python-3-9

PR #34, opened 2023-05-24
chore: add ruff for Python linting

gjreda:
- https://github.com/refstudio/refstudio/issues/26

Adding ruff with its default settings as a pre-push hook for now. I imagine we'll want to tweak some settings later (like line length).

Once this is merged, devs will need to run:
```
$ poetry install
```

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -4,3 +4,4 @@
 yarn lint:check
 # yarn test
 
+ruff src-python
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -4,4 +4,4 @@
 yarn lint:check
 # yarn test
 
-poetry run ruff src-python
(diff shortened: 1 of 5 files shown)


cguedes:
@gjreda Does it make sense add package.json scripts for _check_ and _fix_ like we did for eslint and prettier?

You can check https://github.com/refstudio/refstudio/pull/37 for what I've implemented for prettier.

cguedes (changes requested):
@gjreda Does it make sense add package.json scripts for check and fix like we did for eslint and prettier?

You can check https://github.com/refstudio/refstudio/pull/37 for what I've implemented for prettier.

gjreda:
@cguedes should be good now

cguedes (approved):
Great thanks!

PR #37, opened 2023-05-25
fix: run prettier rules in git `pre-push` hook

### LINKED ISSUE #36: Prettier rules don't run in git pre-push hook
Prettier rules (`--check`) should also be enforced by the push hook.

We should create `package.json` scripts to check and fix prettier issues.

cguedes:
This fixes #36


sehyod (approved):

hammer:
@cguedes what are the other changes in this commit? Are they changes suggested by running `prettier`?

cguedes:
@hammer there are no other changes in this PR.

We created a new command `prettier:check` that was added to the git `pre-push` hook.

image

One can run the `prettier:fix` to fix any prettier issue in the project (ex: missing semi). But that shouldn't happen bc we have format on save configured in the project. But one might save without format or edit outside of vscode so this is an extra safeguard.

cguedes:
>  I presume those are all changes suggested by prettier but just want to be sure that's the case.

You are correct. These are only mandatory format fixes (for the git push to be accepted).

Issue #35, opened 2023-05-25
Vertical scroll within a panel does not work


sehyod:
Vertical scroll does not work within the different components of the app, making it impossible to select some references when the list is too long and leading to some strange behaviour when the text in the editor is too long.

cguedes:
This was fixed by #40

PR #40, opened 2023-05-25
fix: allow flexboxes to scroll

sehyod:
**Problem**
Panel components have an `overflow` property set to `hidden`, that is why adding the tailwind `overflow-scroll` class does not make the views scrollable. #35 
**Solution**
Remove `overflow-scroll` class and make views flexes, with a scrollable child

https://github.com/refstudio/refstudio/assets/58954208/ee89dc39-8969-4a60-8b5c-59f8a1fcc2ea

### DIFF
--- a/src/views/AIView.tsx
+++ b/src/views/AIView.tsx
@@ -10,18 +10,20 @@ export function AIView({ selection }: { selection: string | null }) {
   }, [selection]);
 
   return (
-    <div>
+    <>
       <h1>AI Interactions</h1>
-      <p className="my-4 italic">Select some text in the editor to see it here.</p>
+      <div className='flex-1 overflow-scroll'>
+        <p className="my-4 italic">Select some text in the editor to see it here.</p>
 
-      {selection && (
-        <div className="flex flex-col gap-4">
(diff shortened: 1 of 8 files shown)


src/views/AIView.tsx
```diff
@@ -10,17 +10,19 @@ export function AIView({ selection }: { selection: string | null }) {
   }, [selection]);
 
   return (
-    <div>
+    <div className='flex flex-col h-full w-full'>
       <h1>AI Interactions</h1>
-      <p className="my-4 italic">Select some text in the editor to see it here.</p>
+      <div className='flex-1 overflow-scroll'>
```
cguedes, line 15: Do you think worth creating a wrapper component for a panel view with title so that we can reuse?

sehyod: Yes, I think it makes sense! I will update the PR

sehyod: As discussed offline, only AI Interactions and References could currently use a common wrapper, so creating the wrapper isn't that worth. We should open an issue to refactor the folders view to something more similar to VSCode, with the possibility to drag-and-drop in a specific folder, etc. The wrapper can then be implemented in the PR refactoring the view.

PR #47, opened 2023-05-25
chore: enable more ESLint rules and fix existing violations

danvk:
See #18, follow-on to #24

These are based on rules I've used on projects in the past. It's a mix of style conventions (`curly`), nudges towards good modern JS style (`prefer-destructuring`, `prefer-optional-chain`), pushes towards better TypeScript types and some rules around using Promises correctly. I've also included a rule to sort JSX props which I've found very convenient.

Most of these are auto-fixable.


.eslintrc.cjs
```diff
@@ -0,0 +1,128 @@
+/** @type {import("@types/eslint").Linter.Config} */
+module.exports = {
```
danvk, line 2: I moved this out of `package.json` and into its own file so that we could use things like single quotes and comments.

src/App.tsx
```diff
@@ -27,13 +27,13 @@ function App() {
 
   return (
     <PanelGroup autoSaveId="refstudio" direction="horizontal">
-      <Panel defaultSize={20} collapsible className="overflow-scroll p-4">
+      <Panel className="overflow-scroll p-4" collapsible defaultSize={20}>
```
danvk, line 30: This change comes from `react/jsx-sort-props`.

src/TipTapEditor/CollapsibleBlock/CollapsibleBlock.css
```diff
@@ -1,16 +1,16 @@
 .collapsible-block {
   /* @apply debug; */
-  @apply flex my-2 p-2 pl-0 rounded-lg;
+  @apply my-2 flex rounded-lg p-2 pl-0;
```
danvk, line 3: There's some diff noise here from running `yarn fmt`

src/TipTapEditor/CollapsibleBlock/nodes/CollapsibleBlockNode.ts
```diff
@@ -25,15 +25,14 @@ export const CollapsibleBlockNode = Node.create({
 
   addKeyboardShortcuts() {
     return {
-      'Mod-Enter': () => {
-        return this.editor
+      'Mod-Enter': () =>
+        this.editor
```
danvk, line 29: This is the `arrow-body-style` rule.

src/TipTapEditor/TipTapEditor.tsx
```diff
@@ -11,29 +11,36 @@ import { EDITOR_EXTENSIONS, INITIAL_CONTENT } from './TipTapEditorConfigs';
...
+          onSelectionChange(newEditor.view.state.doc.textBetween(from, to));
+        },
+      }),
+    );
   }, [editorContent, onSelectionChange]);
 
   useEffect(() => {
-    if (!editor) return;
+    if (!editor) {
+      return;
```
danvk, line 29: This is the `curly` rule in action.

src/filesystem.tsx
```diff
@@ -57,14 +57,16 @@ export async function readFile(file: FileEntry) {
 function sortedFileEntries(entries: FileEntry[]): FileEntry[] {
   return entries
     .sort((fileA, fileB) => {
-      if (fileA.children === null) return -1;
-      if (fileB.children === null) return 1;
-      return fileA.name?.localeCompare(fileB.name || '') || 0;
+      if (fileA.children === undefined) {
```
danvk, line 60: This is [@typescript-eslint/no-unnecessary-condition](https://typescript-eslint.io/rules/no-unnecessary-condition) and seems to be a real issue (the comparison should be against `undefined`, not `null`).

src/views/CenterPaneView.tsx
```diff
@@ -17,22 +17,32 @@ export function CenterPaneView({ file, ...props }: CenterPaneViewProps) {
     if (file && isTipTap(file)) {
       setLoading(true);
       (async () => {
-        const content = await readFile(file);
-        const textContent = new TextDecoder('utf-8').decode(content);
+        const newBytes = await readFile(file);
```
danvk, line 20: This was `no-shadow`.

src/views/FoldersView.tsx
```diff
@@ -12,27 +12,38 @@ export function FoldersView({ onClick }: { onClick?: (fileEntry: FileEntry) => v
...
+    readAllProjectFiles()
+      .then((newFiles) => {
+        setFiles(newFiles);
+        const selected = newFiles.find((f) => f.name?.endsWith('.tiptap'));
+        if (selected) {
+          setSelectedFile(selected); // We need this because we might be selecting a DOT_FILE
+          onClick?.(selected);
+        }
+      })
+      .catch((e) => {
```
danvk, line 24: This is [@typescript-eslint/no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises).

danvk (commented):
Some additional rules for consideration.

.eslintrc.cjs
cguedes: I don't mind about this. The file is getting very big and we can take advantage of single quotes and comments.
What do you think @sergioramos?

src/App.tsx
cguedes: If this is auto-sortable on save it's ok for me.

src/TipTapEditor/CollapsibleBlock/CollapsibleBlock.css
cguedes: Somehow the auto-sort for did not pick up css files.

.eslintrc.cjs
sergioramos: The reason I suggested in `package.json` was to not have tons of root files. But I'm OK either way.

src/TipTapEditor/TipTapEditor.tsx
cguedes: I do like to this type of "fast-return" conditions being a single line so that we can read the "pre-conditions" at the top without any line breaks.

Sometimes you have 2 or 3 (unrelated) conditions you need to check (ex: with some ! expressions). IMO it's a good practice to keep them different statements (ex: for diff and legibility).

With this rule we would have 3 rules x 3 lines = 9 lines of code instead of only 3 like the example below.

```javascript
useEffect(() => {
    if (!editor) {
      return
    }
    if (draft) {
      return
    }
    if(anotherRule & anotherRule.field) {
      return
   }
   // ...
 }, [...])
```

vs

```js
useEffect(() => {
    if (!editor) return
    if (draft) return
    if(anotherRule & anotherRule.field) return
    // ...
 }, [...])

src/TipTapEditor/CollapsibleBlock/nodes/CollapsibleBlockNode.ts
sergioramos: Don't you prefer `block body` for arrow functions when it has a line break? I find it easier to read, and I always think about a well-known security issue a few years ago because an if/else didn't have curly braces and someone added another instruction (so, another line) which would run every time (regardless of the condition)

cguedes (approved):

src/TipTapEditor/CollapsibleBlock/nodes/CollapsibleBlockNode.ts
danvk: I don't mind this, actually! The if/else continuation line issue is very real, which is why I think `curly` is a good idea. But since the thing on the RHS of `=>` is an expression, not a statement, I think there's less room for trouble. In most cases if you added an extra statement after an arrow function with a line break, it would be a syntax error or create dead code (which TypeScript would complain about).

src/TipTapEditor/TipTapEditor.tsx
danvk: If you guys would like to set `curly: multi-line` (https://eslint.org/docs/latest/rules/curly#multi-line) that's fine by me. My preference would be slightly more strict than that rule, though. If you have an `else` or an `else if`, you should be using braces. This is definitely a slippery slope, see @sergioramos's comment above about possible security issues.

Issue #46, opened 2023-05-25
Setup testing and coverage reports for Python


gjreda:
Can use pytest and one of its plugins for coverage reporting (likely `pytest-cov`).

PR #51, opened 2023-05-26
Adding scaffolding for pytest and coverage reporting

gjreda:
- https://github.com/refstudio/refstudio/issues/46


cguedes (approved):

sehyod (dismissed):
Thank you for that! I will work on adding this to the CI workflow as soon as this branch has been merged.
When trying the branch locally, I noticed it created a few files, in `__pycache__` folders, and the resulting .coverage file. Would it make sense to add these to .gitignore so we don't accidentally push them? Or do we want to have them in the repo?

Issue #60, opened 2023-05-29
AI interactions: rewrite


hammer:
Some operations on a span or selection:

- Change tone (e.g. casual, formal)
- Change length
- Change tense
- Change form

[RewriteLM: An Instruction-Tuned Large Language Model for Text Rewriting](https://arxiv.org/abs/2305.15685) (2023) is a recent Google paper that proposes a benchmark, OpenRewriteEval, that has some useful examples.

gjreda:
Pulling over relevant API design notes from #19 

### Span Operation: Rewrite a highlighted block of text (concisely)

**Action**:
- User highlights block of text and there is some UX component that allows them to ask for the text to be rewritten in a more concise manner.
- There will likely be other types of "rewrite" operations beyond just conciseness, but conciseness can be a starting point.
- UX for this is TBD

**Input**:
- Highlighted span of text (for now ... I can imagine the prior blocks of text or entire document would also be useful here)

**Sidecar Response**:
- json, list of candidate rewrites

Example Response:
```json
[
  {
    "id": 1,
    "text": "I'm a paragraph"
  },
  {
    "id": 2,
    "text": "I'm an another option"
  }
]
```

Issue #57, opened 2023-05-27
Free hosted models


hammer:
It would be nice for testing and development if we could find an alternative to OpenAI APIs that is free at low volumes.

Some ideas that I have not yet dug into:

- [Hugging Face Inference API](https://huggingface.co/docs/api-inference/index)
- [MosaicML Inference](https://www.mosaicml.com/inference)
- [Replicate](https://replicate.com/pricing) gives some free credits but it doesn't say how much (?)
- [Modal](https://modal.com/pricing) gives $10/mo in free credits
- [Baseten](https://www.baseten.co/pricing) gives $30 in free credits
- [Modelbit](https://www.modelbit.com/) does not make pricing easily discoverable
- [Lambda Demos](https://cloud.lambdalabs.com/demos): don't seem to have APIs

gjreda:
Spent some time trying a few models via HF's Inference API. My takeaways were:
1. If the model is sufficiently sized (> 10GB), HF requires you to use their commercial spaces or standup your own paid [inference endpoint](https://huggingface.co/inference-endpoints), ultimately putting you back into metered billing territory.
2. The smaller, freely available models are generally task specific (e.g. [BART](https://huggingface.co/facebook/bart-large-cnn) for summarization, [roberta](https://huggingface.co/deepset/roberta-base-squad2) for Q&A), so we'd have to plan for multiple models, rather than just a single LLM acting as a sufficiently general API for all tasks.

gjreda:
Some pricing notes:

**BaseTen**
Deploying Flan-T5 XL on BaseTen currently costs $0.04040 per minute while running. BaseTen will scale the model down to zero after one hour without any requests.

**OpenAI**
When building the ["chat with PDFs" prototype](https://github.com/gjreda/scratch-pdf-bot), I spent ~$2.00 in API calls to GPT-3.5 and embeddings creation. Since this was a prototype, I opted for ease of development and made no effort to be smart about calling the API (caching).

<img width="758" alt="image" src="https://github.com/refstudio/refstudio/assets/921995/3f30d682-ae75-4d4b-8e54-985a2666a5dc">

hammer:
Modal has a quantized 4-bit version of Falcon-40B running at https://modal-labs--example-falcon-gptq-get.modal.run. It's too slow to be useful for now, I think. Instruction for self hosting are at https://modal.com/docs/guide/ex/falcon_gptq in case that ever becomes interesting to us.

hammer:
[...]

PR #66, opened 2023-06-01
feat: Add AI interaction for text rewrite

gjreda:
- https://github.com/refstudio/refstudio/issues/60

Adds initial "rewrite" capability to the sidecar. For a first pass, we're only focusing on the ability to rewrite a highlighted block of text in a more concise manner.

After trying some of the model options mentioned in https://github.com/refstudio/refstudio/issues/57, I decided to move forward with OpenAI for now. Ultimately, it felt like there's A LOT of noise around the other various model alternatives and I was starting to feel like I had fallen down a deep dark rabbit hole. When a clear winner emerges for us as a better option, I don't think swapping over to it will be too bad.

Another note is that I've opted to use GPT-3.5 rather than davinci. I felt that GPT-3.5 was producing better, more human-like responses. It also [costs 1/10th of the price](https://openai.com/pricing#language-models) for now.

### Example usage
```bash
❯ poetry run python python/main.py rewrite --text "In the cycling world, power meters are the typical way to objectively measure performance. Your speed is dependent on a lot of factors, like wind, road surface, and elevation. One's heart rate is largely a function of genetics, but also things like temperature. The amount of power you are outputting to the pedals though, is a direct measure of how much work you're doing, regardless of wind, elevation, your body weight, etc."

[{"index": 0, "text": "Power meters are the standard method for objectively measuring cycling performance, as they directly measure the amount of work being done regardless of external factors such as wind, elevation, and body weight."}]
```
Note that I'm passing `index` as a key rather than `id` to maintain consistency with how the OpenAI response refers to these fields. I think we might want to reserve what they call ID for later, when we eventually get to chat interactions.


hammer:
One thought I had on this PR that I forgot to mention in our call: we should probably add a configuration UX to allow users to specify their `OPENAI_API_KEY`, and possibly to configure other aspects of the app. cc @cguedes

gjreda:
Ran a couple of examples using the ML scholarly papers we've been testing against and results still seem pretty good.

Moving this into "ready for review" as I think it makes sense to merge in the initial backend work as one PR and then add in the front-end interaction once we have the UI ready. cc @cguedes @sehyod Merging this in shouldn't cause any issues since it's not called anywhere.

```bash
❯ poetry run python python/main.py rewrite --text "A common misconception about overfitting is that it is caused by noise, like training examples labeled with the wrong class. This can indeed aggravate overfitting, by making the learner draw a capricious frontier to keep those examples on what it thinks is the right side. But severe overfitting can occur even in the absence of noise. For instance, suppose we learn a Boolean classifier that is just the disjunction of the examples labeled “true” in the training set. (In other words, the classi- fier is a Boolean formula in disjunctive normal form, where each term is the conjunction of the feature values of one specific training example). This classifier gets all the train- ing examples right and every positive test example wrong, regardless of whether the training data is noisy or not." | jq
[
  {
    "index": 0,
    "text": "Overfitting is not solely caused by noise, as it can occur even without it. For example, a Boolean classifier that is just the disjunction of the examples labeled \"true\" in the training set can lead to severe overfitting, getting all training examples right but every positive test example wrong."
  }
]
```

cguedes:
> One thought I had on this PR that I forgot to mention in our call: we should probably add a configuration UX to allow users to specify their `OPENAI_API_KEY`, and possibly to configure other aspects of the app. cc @cguedes

Created issue for that new feature: https://github.com/refstudio/refstudio/issues/76

cguedes (approved):

cguedes (approved):
NOTE: I've only checked the code. I didn't had an API key to test.

python/sidecar/interact.py
```diff
@@ -0,0 +1,53 @@
+import json
+import os
+import sys
+
+import openai
+from dotenv import load_dotenv
+from sidecar import prompts
+
+load_dotenv()
+openai.api_key = os.environ.get("OPENAI_API_KEY")
```
cguedes, line 10: @gjreda I've changed this from index notation to `.get(...)` to fix the tests.

I'm not sure if this is the best approach but I feed that we shouldn't depend on this ENV var in the tests.

hammer: I've been thinking about how to test our code when we depend on OpenAI APIs myself, so this PR is a good reminder to open an issue: https://github.com/refstudio/refstudio/issues/88. Let's put thoughts there?

gjreda: @cguedes I think either works. We are not using the API key at all in tests, since we are [mocking the call and response](https://github.com/refstudio/refstudio/pull/66/files#diff-bf83c0d6ab5fabad80921af7c7ac063d3524bb464bb2db06e5fbe240916d2cf7R7-R31). The OpenAI API never actually gets called.

Without the `get`, if the env variable is missing, an exception would get raised once the package was loaded. With the `get`, an exception will get raised once we try to call the API and realize the key is missing. Either way, it shouldn't have an impact on the tests.

gjreda: Said differently, in testing, we could set the env variable to `test` and it shouldn't make a difference.

cguedes (approved):

cguedes (dismissed):

python/sidecar/interact.py
```diff
@@ -0,0 +1,53 @@
...
+            choices.append(
+                {'index': choice["index"], 'text': choice["message"]["content"]}
+            )
+        return choices
+    
+    def call_model(self, messages: list):
+        response = openai.ChatCompletion.create(
+            model=os.environ["OPENAI_CHAT_MODEL"],
+            messages=messages,
+            n=1,  # number of completions to generate
```
sehyod, line 43: Shouldn't it be `n=self.n_options,` here?

src/panels/AIPanel.tsx
```diff
@@ -8,19 +8,29 @@ import { PanelSection } from '../components/PanelSection';
...
-      .then((num) => setReply(String(num)))
-      .catch(console.error);
-  }, [selection]);
+    setRewriteReply('');
+  }, [debouncedSelection]);
+
+  const [rewriteReply, setRewriteReply] = useState('');
+  const [loadingRewrite, setLoadingRewrite] = useState(false);
+
+  function handleRewriteClick() {
```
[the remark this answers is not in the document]
cguedes: @sehyod  not sure is a fit for the hook. Please try and let me know.

python/sidecar/interact.py
cguedes: I think that for now we are only asking for 1 option. But we can use self.n_options that default to 1.

pyproject.toml
```diff
@@ -9,6 +9,10 @@ readme = "README.md"
 python = "^3.9, <3.12"  # pyinstaller requires Python <3.12
 grobid-tei-xml = "^0.1.3"
 grobid-client-python = "^0.0.5"
+sentence-transformers = "^2.2.2"
+lancedb = "^0.1.3"
```
hammer, line 13: We can remove these from this commit, I believe?

hammer:
I have put my OpenAI API key into the environment but I'm still not seeing output when running this PR. It says "Processing..." when I hit "REWRITE" but no text is returned.

How can I get more diagnostics to figure out what's going on?

pyproject.toml
gjreda: Yep. I think this only ended up here because other PRs got merged.

hammer:
I was able to get this working on a fresh checkout. The AI is alive!!!

hammer:
What’s blocking this from going in?

sehyod:
The missing piece here is actually rewriting the selected text with what the AI returns. I think @cguedes was working on it but he is off today. I think we could merge this as is and add the actual rewrite action in a follow up PR

hammer:
Yeah I think there are UX issues to work through on how to take the text returned by the model and incorporate it into the document. Let's just get this in and open up a follow-on issue for what to do with the model-generated text.

gjreda:
Only thing blocking from my perspective is an approving review. I can't merge without it.

sehyod (approved):

Issue #10, opened 2023-05-15
CI/CD plans


hammer:
We'd like to get CI/CD set up quickly for this project. @sergioramos notes that for MacOS testing we'll need to sign our code, as noted in the Tauri docs: https://tauri.app/v1/guides/distribution/sign-macos.

@jonaskruckenberg @denjell-crabnebula any other tips for us on CI/CD for Tauri apps? Obviously there's the testing section in your docs but if there are any practical lessons you've learned that haven't made it into the docs yet we'd love to hear them! In particular we're concerned about testing on Windows since we don't have any Windows users. Any pro tips on how to build apps to make them maximally compatible across Mac and Windows?

sergioramos:
On every Commit/PR:

<details>

<summary>Example</summary>

```yaml
name: Continuous Integration

on:
  push:
    branches:
      - master
      - main
  pull_request:
```

</details>



- [ ] Install dependencies and setup environment (with caches)

<details>

<summary>Example</summary>

```yml
  setup:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3.5.2
      - name: Setup Node.js
        uses: actions/setup-node@v3.6.0
        with:
          node-version: 'lts/*'
          cache: 'yarn'
      - name: Install dependencies
        run: yarn install --immutable
        env:
          YARN_CHECKSUM_BEHAVIOR: update
          YARN_ENABLE_SCRIPTS: 0
```

</details>

- [ ] Lint commits

<details>

<summary>Example</summary>

```yml
  commitlint:
    needs: setup
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3.5.2
        with:
          fetch-depth: 0
      - name: Commit Lint
        uses: wagoid/commitlint-github-action@v5.4.1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

</details>

- [ ] Eslint

<details>

<summary>Example</summary>

```yml
  eslint:
    needs: setup
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3.5.2
      - name: Setup Node.js
        uses: actions/setup-node@v3.6.0
        with:
          node-version: 'lts/*'
          cache: 'yarn'
      - name: Install dependencies
        run: yarn install --immutable
      - name: ESLint
        run: yarn run lint
```

</details>

[...]

PR #67, opened 2023-06-01
chore: check typescript error on push

cguedes:
Add a new yarn command to check typescript errors.

```bash
yarn ts:check
```

This command/check was also added to the git push hook to ensure that we can't push code that don't compile.

Also updates the `yarn check` to include this check.

Related to #24 and #10 

@sehyod


.husky/pre-push
```diff
@@ -3,6 +3,7 @@
 
 yarn lint:check
```
danvk, line 4: Can this just run `yarn check`? What's the distinction here?

cguedes: I was wondering exactly that.
In a way we have more control (and legibility) over what get executed on push, but we can also forget to add them here!

So I'm ok to update this to use `yarn check` instead.

What do you think @sehyod ?

sehyod: I agree with using `yarn check` in the pre-push hook: it's better not to have to update the code in multiple places every time we want to add a new check

danvk (approved):

cguedes:
@sergioramos this is ready to be merged.

Issue #56, opened 2023-05-26
Sidecar dylib binary errors: PyInstaller --onedir does not play nicely with Tauri sidecar pattern


gjreda:
The sidecar binary will inconsistently throw the following error:
```
usage: main [-h] [--debug] {ingest,ai} ...
main: error: argument command: invalid choice: 'from multiprocessing.resource_tracker import main;main(14)' (choose from 'ingest', 'ai')
```
Sometimes this error will not be thrown. Other times it will be thrown and not return any results. Other times it will return the proper results _and then_ throw this error.

It seems related to spawning multiple processes deeper within the binary, likely coming from one of the `sentence-transformers` dependencies. We do not currently do any explicit multiprocessing.

This seems like a relevant issue from the pyinstaller github:
- https://github.com/pyinstaller/pyinstaller/issues/4865

gjreda:
For PR #29, I toyed with removing `--onefile` and it seemed to result in different errors. Some examples:

```
[Error] Unhandled Promise Rejection: Error: Traceback (most recent call last):
  File "PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py", line 110, in <module>
  File "PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py", line 97, i...

	(anonymous function) (filesystem.tsx:50)

[Error] Unhandled Promise Rejection: Error: Traceback (most recent call last):
  File "PyInstaller/hooks/rthooks/pyi_rth_pkgres.py", line 16, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 352, in exec_module
  File ...

	(anonymous function) (filesystem.tsx:50)
```

This is coming from the JS console, so I don't think the Unhandled Promise is actually related ... the error is happening in the Python binary and propagating up.

I'll plan to investigate this further in conjunction with https://github.com/refstudio/refstudio/issues/48

gjreda:
Investigating this further ...

Running PyInstaller in `--onedir` mode (as opposed to `--onefile`):
```bash
$ poetry run pyinstaller --noconfirm --hidden-import torch --collect-data torch --recursive-copy-metadata torch --hidden-import pyarrow --collect-data pyarrow --recursive-copy-metadata pyarrow --collect-binaries pyarrow --collect-submodules pyarrow --recursive-copy-metadata tqdm --recursive-copy-metadata sentence-transformers --distpath src-tauri/bin/python src-python/main.py
```

This produces a binary that works on its own:
[...]

Issue #48, opened 2023-05-25
Sidecar process profiling


gjreda:
Since adding `sentence-transformers` and `lancedb`, simple calls to the sidecar (e.g. `get_word_count`) are significantly slower. Do profiling to understand why.

I have a suspicion its related to requiring `torch` as a `--hidden-import` when building the binary.

Here is the command to build the binary:
```bash
poetry run pyinstaller --onefile --noconfirm --hidden-import torch --collect-data torch --recursive-copy-metadata torch --hidden-import pyarrow --collect-data pyarrow --recursive-copy-metadata pyarrow --collect-binaries pyarrow --collect-submodules pyarrow --recursive-copy-metadata importlib_metadata --recursive-copy-metadata tqdm --recursive-copy-metadata sentence-transformers --distpath src-tauri/bin/python src-python/main.py
```

And here is the PyInstaller version:
```bash
$ poetry run pyinstaller --version
5.11.0
```

gjreda:
Calling the binary version of `ingest` _on an empty directory_ suggests this is likely related to startup.
```bash
$ time ./src-tauri/bin/python/main-aarch64-apple-darwin/main ingest --pdf_directory="/Users/greg/Library/Application Support/com.tauri.dev/project-x/uploads"
{"project_name": "project-x", "references": []}

./src-tauri/bin/python/main-aarch64-apple-darwin/main ingest   7.35s user 4.65s system 60% cpu 19.946 total
```

The same command on the source `main.py`, on the same empty directory:
```bash
$ time poetry run python src-python/main.py ingest --pdf_directory="/Users/greg/Library/Application Support/com.tauri.dev/project-x/uploads"
{"project_name": "project-x", "references": []}

poetry run python src-python/main.py ingest   1.63s user 0.85s system 374% cpu 0.663 total
```

gjreda:
There's some discussion here that suggests it could be related to our use of the `--onefile` flag: https://github.com/pyinstaller/pyinstaller/issues/7060

Rebuilding the binary without `--onefile`:
```bash
[...]

PR #74, opened 2023-06-01
feat: use python binary as a resource instead of a sidecar

gjreda:
- https://github.com/refstudio/refstudio/issues/56
- https://github.com/refstudio/refstudio/issues/48

This PR no longer creates the python binary in `--onefile` mode, which is why the binary is so slow, and instead we now use the default `--onedir` mode.

For this to work with Tauri, we cannot bundle the binary as a sidecar and instead need to add it as a resource.

The first call to the binary is still fairly slow, but subsequent calls are much quicker. We can probably warm up the binary on app startup with a simple "hello world" command.

**Open Questions**
- [x] Do we need to change how the `interactWithAi` command fires, so that we aren't firing so many of them?
- [x] Do we have to include the target-triple in the filepath? How do we bundle resources for other OS using this pattern? I was only able to get this working by specifying the full path, including the target-triple.
- [x] Do we need to declare multiple `scopes` even if the underlying cmd will be the same?

https://github.com/refstudio/refstudio/assets/921995/f8784760-30a6-45da-bc7e-c0dc197f430b


gjreda:
@cguedes @sehyod Can y'all take a look at the video above and front-end changes in this PR? Do we need to change something about how `interactWithAi` with fires now? So that we aren't firing this command too quickly and getting out of order results?

hammer:
Nice!

I was imagining a world where we ship the binary as a sidecar but the shared libraries as resources. With this approach do we lose Tauri’s process supervision? In other words do we need to clean up the resources used by our Python process such as sockets, and do we need to monitor and restart it ourselves if it fails?

Also I agree we should not fire AI interactions on select, we should have a way for the user to trigger the interactions either in the editor or the sidebar or both.

gjreda:
> With this approach do we lose Tauri’s process supervision? In other words do we need to clean up the resources used by our Python process such as sockets, and do we need to monitor and restart it ourselves if it fails?

I don't believe so since we're ultimately still using the same [command api](https://tauri.app/v1/api/js/shell#command), but agree we should get some clarity on the tradeoffs between this approach and the sidecar.

cguedes (approved):

PR #80, opened 2023-06-02
Fix case sensitive components folder name

sergioramos:


cguedes (approved):

cguedes:
Thanks @sergioramos 🙏

Issue #75, opened 2023-06-02
Create an internal FileEntry type


cguedes:
The code is already using `@tauri-apps/api/fs.FileEntry` in 7 files (filesystem, openFilesState, multiple UI components).

The existing interface of FileEntry also lacks:
  - separation (types) between file and directory
      - we need to check if children is not undefined
  - extension name for files
  - additional metadata about the file

I propose that we create an internal type for this and make the `filesystem.ts` the integration point with Tauri's API.

Inspiration for filesystem types and names:
  * https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e858f398c44ef759a64dd49854fd3470e6b65731/types/filesystem/index.d.ts#LL112C11-L112C26

>  I think we won't get much with that change (we don't have `extname` available for a file). This code (i.e. checking for the extension) would be simplified with the creation of a custom type for FileEntry (derived from Tauri's FileEntry).

_Originally posted by @cguedes in https://github.com/refstudio/refstudio/pull/62#discussion_r1214216652_

PR #81, opened 2023-06-02
feat: populate and suggest references after PDF ingestion

### LINKED ISSUE #49: Populate new references in ReferencesView after PDF ingestion
After new PDFs are ingested, update the references list found in ReferencesView with metadata about the new references.

sehyod:
This PR closes #49.
It adds the following features:
- Showing a spinner in the references view while uploaded files are being uploaded
- Updating the references view with new references once ingestion is done
- Changing the way references are represented in the editor to use the [markdown syntax](https://quarto.org/docs/authoring/footnotes-and-citations.html#sec-citations)
- Connecting the suggestion widget to the actual list of references so that typing `[` in the editor enables you to insert a reference

Screenshot 2023-06-02 at 17 49 03
Screenshot 2023-06-02 at 17 50 28

Implementation choices:
- One jotai atoms has been created to store references
- When ingestion is running, file uploading is disabled
- This PR uses the [Mention extension](https://tiptap.dev/api/nodes/mention) of TipTap to open the suggestion list when typing `[`
  - This works for now but the extension is not really customisable and can only be rendered as text, we may have to create our own custom extension if we want to support having a tooltip with information about the reference when hovering it in the editor
- This PR uses [`react-popper`](https://github.com/floating-ui/react-popper) to render the suggestion list
  - The `suggestion` extension requires a library to position the list
  - `Mention` comes with `tippy` but we need a react implementation because jotai atoms are only available in the React world (if we want to avoid having to handle stores)
  - [`tippyjs/react`](https://github.com/atomiks/tippyjs-react) would be an alternative we can use


package.json
```diff
@@ -42,6 +42,7 @@
     "react-drag-drop-files": "^2.3.10",
     "react-icons": "^4.8.0",
     "react-pdf": "^6.2.2",
+    "react-popper": "^2.3.0",
```
sergioramos, line 45: Doesn't ProseMirror support autocomplete? If so, could we use it for this use-case?

src/TipTapEditor/ReferenceNode/ReferencesList.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+import { getReferencesAtom } from '../../atoms/referencesState';
+import { cx } from '../../cx';
+import { ReferenceItem } from '../../types/ReferenceItem';
+
+export type ReferenceListProps = SuggestionProps<{ id: string; label: string }>;
+
+export const ReferencesList = forwardRef((props: ReferenceListProps, ref) => {
+  const { command, clientRect, query } = props;
+  console.log(clientRect?.());
```
cguedes, line 17: You might want to remove this.

src/atoms/referencesState.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { ReferenceItem } from '../types/ReferenceItem';
+
+const pdfIngestionProcessAtom = atom<string[]>([]); // ids of pending processes for now, this could be a record with progress percentage in the future
+
+export const isPdfIngestionRunningAtom = atom((get) => get(pdfIngestionProcessAtom).length > 0);
+
+export const addPdfIngestionProcessAtom = atom(null, (_get, set, value: string) => {
+  set(pdfIngestionProcessAtom, (pendingProcesses) => [...pendingProcesses, value]);
+});
```
cguedes, line 11: Given that this logic is so simple (push to array), shouldn't one use the `pdfIngestionProcessAtom` instead?

src/atoms/referencesState.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const removePdfIngestionProcessAtom = atom(null, (_get, set, value: string) => {
+  set(pdfIngestionProcessAtom, (pendingProcesses) => pendingProcesses.filter((processId) => processId !== value));
+});
```
cguedes, line 15: Same for the removal, shouldn't one use only `pdfIngestionProcessAtom`

src/atoms/referencesState.ts
```diff
@@ -0,0 +1,31 @@
...
+
+type ReferencesState = Record<string, ReferenceItem>;
+
+const referencesAtom = atom<ReferencesState>({});
+
+export const getReferencesAtom = atom((get) => Object.values(get(referencesAtom)));
+
+export const getDerivedReferenceAtom = (referenceId: string) => atom((get) => get(referencesAtom)[referenceId]);
```
cguedes, line 23: Isn't this a use case for the `atomFamily`?

src/atoms/referencesState.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const setReferencesAtom = atom(null, (_get, set, references: ReferenceItem[]) => {
+  const updatedReferences = {} as ReferencesState;
+  references.forEach((reference) => {
+    updatedReferences[reference.id] = reference;
+  });
+  set(referencesAtom, updatedReferences);
+});
```
cguedes, line 31: I'm a bit confused by this write-only atom.
If the idea is to set the `referencesAtom` why can't one use the atom directly instead of via a write-only atom.

Maybe we want to read the existing `referencesAtom` and use that as initial value (line 26). Then this would be a _patial update_ for the references atoms.

src/filesystem.tsx
```diff
@@ -88,3 +90,11 @@ function sortedFileEntries(entries: FileEntry[]): FileEntry[] {
...
     }));
 }
+
+function parsePdfIngestionResponse(response: PdfIngestionResponse): ReferenceItem[] {
+  return response.references.map((reference) => ({
+    id: reference.filename_md5,
+    title: reference.title ?? reference.source_filename.replace('.pdf', ''),
+    authors: reference.authors.map((author) => ({ fullName: author.full_name, surname: author.surname })),
+  }));
+}
```
cguedes, line 100: We should move the PDF Ingestion related code to outside of the `filesystem.tsx`

src/types/PdfIngestion.ts
```diff
@@ -0,0 +1,18 @@
+export interface PdfIngestionResponse {
```
cguedes, line 1: I would move this to the same new file/folder as we would move the pdf ingestion methods in filesystem.ts

cguedes (changes requested):

src/atoms/referencesState.ts
sehyod: Yes, the logic is simple for now, but I had in mind that I could get more complex in the future, like if we want to store a progress percentage along with the progress. This atom is the atom allowing a component to say "An ingestion progress has started" without knowing anything about the underlying representation we have chosen to store processes.
I agree in this case this may be overcomplicating and we can actually just expose the `pdfIngestionProcessAtom`

src/atoms/referencesState.ts
sehyod: Same as before, the idea was to decouple the atom to say that a process has ended, from the actual representation of processes.

src/atoms/referencesState.ts
sehyod: I agree this could be `addReferencesAtom`, so we keep the current state as default value and add the new values. I did it this way because the python process returns everything anyway, but that is a bit weird, I will update it.

About the write-only atom is because the internal way of storing references is different from the way components use them: it is convenient to have an object to store them, so that we can get any reference by id in O(1) (or maybe this should be a `Map`, I'm happy to change it), but then the components handle arrays of references, not objects. When the ingestion process is done, it returns an array, then it shouldn't be the component's job to transform this array into the structure we use in the atom, hence this write-only atom. Here again, decoupling internal storage from getter/setter allows us to change the internal structure any time without having to update the code everywhere.

src/filesystem.tsx
sehyod: I definitely agree, this will be done as part of #75 rather than in this PR

src/TipTapEditor/ReferenceNode/ReferencesList.tsx
sehyod: Nice catch, thank you!

cguedes (approved):

PR #63, opened 2023-05-30
ci: add workflows

sehyod:
Fixes #10

PR #83, opened 2023-06-02
fix: remove embedding creation from pdf ingestion

### LINKED ISSUE #50: Remove embedding creation from PDF ingestion
We want to populate the ReferenceView with metadata once the uploaded references have been parsed: https://github.com/refstudio/refstudio/issues/49

As currently constructed, Reference metadata is not returned until the entire pipeline runs. However, metadata is available as json fairly early on in the pipeline.

Separating embedding creation from the rest of the ingestion pipeline would allow us to return the metadata response sooner and allow things to feel snappier.

gjreda:
Fixes #50

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "lint:fix": "eslint --fix --ext .js,.ts,.tsx ./src",
     "ts:check": "tsc --noEmit",
     "preview": "vite preview",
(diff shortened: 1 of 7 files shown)


gjreda:
@sehyod I think this should resolve the issues you're seeing in https://github.com/refstudio/refstudio/pull/63. I think it'd be good to get CI/CD working properly and then I can worry about adding this back in later, since we do not need it right now.

PR #87, opened 2023-06-05
Ensure main.spec exists before removing

cguedes:
Also don't remove all *.spec files bc we might be removing files we don't want

### DIFF
--- a/scripts/remove_binary.sh
+++ b/scripts/remove_binary.sh
@@ -3,4 +3,6 @@
 # Remove binary files from the repository
 rm -rf src-tauri/bin/python
 rm -rf build
-rm *.spec
+
+# Ensure main.spec exists before removing
+touch main.spec && rm main.spec


cguedes:
CC @gjreda

gjreda (approved):

hammer:
Is `rm -f main.spec` not preferred? I don't really care just curious if there's a reason to prefer the `touch` approach here...

cguedes:
> Is `rm -f main.spec` not preferred? I don't really care just curious if there's a reason to prefer the `touch` approach here...

@hammer the original code was using a wildcard `rm *.spec` that don't work with `-f`. We need to use `rm -f "*.spec"` if we use the `*`. 

Now that we only remove the `main.spec` file we can use `rm -f main.spec`. Will add this fix to #99 (also related with scripts)

PR #92, opened 2023-06-05
refactor: use internal FileEntry type

### LINKED ISSUE #75: Create an internal FileEntry type

sehyod:
Fixes #75 
** Problem **
Our code should not rely on the tauri `FileEntry` type. Moreover, this type lacks some useful fields like `isFolder`
** Solution **
Create our own internal `FileEntry` type with a discriminating union. The tauri type should only appear in the interface (`filesystem.ts`), which then converts it to our internal type, which is used acrossed the rest of the app

### DIFF
--- a/src/components/FileTree.tsx
+++ b/src/components/FileTree.tsx
@@ -1,7 +1,7 @@
-import { FileEntry } from '@tauri-apps/api/fs';
 import { VscFile, VscFolder } from 'react-icons/vsc';
 
 import { cx } from '../cx';
+import { FileEntry } from '../types/FileEntry';
 
 interface FileTreeBaseProps {
   root?: boolean;
@@ -26,10 +26,11 @@ function FileTreeNode({ file, onClick, rightAction, selectedFiles }: FileTreeBas
   if (!file) {
     return null;
   }
-  const isFolder = !!file.children;
+
+  const { isFolder } = file;
 
   // Hide DOT_FILES
-  if (file.name?.startsWith('.')) {
+  if (file.name.startsWith('.')) {
     return null;
   }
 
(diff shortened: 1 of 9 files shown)


src/components/FileTree.tsx
```diff
@@ -26,10 +26,11 @@ function FileTreeNode({ file, onClick, rightAction, selectedFiles }: FileTreeBas
   if (!file) {
     return null;
   }
-  const isFolder = !!file.children;
+
+  const { isFolder } = file;
 
   // Hide DOT_FILES
-  if (file.name?.startsWith('.')) {
+  if (file.name.startsWith('.')) {
```
cguedes, line 33: Maybe we can also have a flag for this text

`isDotfile: boolean`  // note that this should exist in both file and directory.

cguedes (changes requested):
Just added a note about the dotfiles

cguedes (approved):

cguedes (approved):

PR #93, opened 2023-06-05
68-collapsible-left-and-right-sidebars

cguedes:


src/App.tsx
```diff
@@ -12,36 +13,93 @@ import { PdfViewerAPI } from './types/PdfViewerAPI';
...
 
-      <Panel defaultSize={60}>
-        <MainPanel editorRef={editorRef} pdfViewerRef={pdfViewerRef} />
-      </Panel>
+function RightPanelWrapper() {
+  const [closed, setClosed] = React.useState(false);
 
+  if (closed) {
+    return (
+      <div className="absolute bottom-0 right-0 flex border border-slate-300 bg-slate-100 px-4 py-2">
```
sehyod, line 85: Just a comment about the UX: it feels a bit weird to me to have to click on the chevron to reopen the panel. It would feel more natural to be able to click anywhere on this div to reopen. What do you think?

src/components/PanelWrapper.tsx
```diff
@@ -1,11 +1,22 @@
...
 
-export function PanelWrapper({ title, children }: { title: string; children: React.ReactNode }) {
+export function PanelWrapper({
+  title,
+  close = false,
+  onCloseClick,
+  children,
+}: {
+  title: string;
+  close?: boolean;
```
sehyod, line 11: I think this name is a bit misleading. Shouldn't it be something like `closable` instead?

sehyod (dismissed):

src/components/PanelWrapper.tsx
cguedes: ok. Fair point. Will update this prop.

src/App.tsx
sehyod: Yes, perfect! I agree the icon should stay

sehyod (approved):

hammer:
@cguedes were the issue I found fixed?

cguedes:
@hammer not yet. Will work on this now.

Issue #79, opened 2023-06-02
PDF ingest does not work properly with non-scholarly documents


gjreda:
- https://ref-studio.slack.com/archives/C05525DV9HN/p1685694897585369?thread_ts=1685633929.177849&cid=C05525DV9HN

From @sehyod 

> I did some testing this morning and it appears there is indeed an issue with one of the pdfs I had in my directory. I'm joining the pdf to this message but it's not a scholar pdf, just a text pdf I used while working on the PDF viewer.

Since it is not a scholarly PDF, I suspect that grobid is unable to parse it and then downstream processing within `ingest.py` is not handling that properly.

sehyod:
To give more context, the pdf I was using was a lorem-ipsum pdf, containing only text. I think the file has been created with Word, this may be where the issue come from.
When running the ingestion, the sidecar doesn't return anything (the stdout is empty) and the error trace is the following:
```
Traceback (most recent call last): (filesystem.tsx, line 52)
  File "main.py", line 48, in <module>
  File "sidecar/ingest.py", line 245, in main
  File "sidecar/ingest.py", line 53, in run
  File "sidecar/ingest.py", line 232, in create_and_persist_embeddings
  File "lancedb/db.py", line 97, in create_table
  File "lancedb/table.py", line 216, in create
  File "lancedb/table.py", line 36, in _sanitize_data
  File "lancedb/table.py", line 242, in _sanitize_schema
  File "lancedb/table.py", line 256, in _sanitize_vector_column
ValueError: 'vector' is not in list
[85999] Failed to execute script 'main' due to unhandled exception!
usage: main [-h] [--debug] {ingest,ai} ...
main: error: argument command: invalid choice: 'from multiprocessing.resource_tracker import main;main(23)' (choose from 'ingest', 'ai')
```

gjreda:
Some more notes from Slack:
> Yes, I am starting with a clean directory (I have run the reset_uploads script), I upload a "regular" PDF (Hidden technical debt...) and when I log the stderr in the runPDFIngestion function, here is the result:
usage: main [-h] [--debug] {ingest,ai} ... (filesystem.tsx, line 52)
main: error: argument command: invalid choice: 'from multiprocessing.resource_tracker import main;main(23)' (choose from 'ingest', 'ai')

> But everything else is working as expected and I do get the correct stdout. So if I hadn't been logging the stdout I wouldn't have noticed there was an issue (edited)

sehyod:
[...]

PR #103, opened 2023-06-06
fix: PDF Ingestion bug when Grobid is unable to parse the reference PDF

gjreda:
- https://github.com/refstudio/refstudio/issues/79

This fixes two underlying bugs:
1. When Grobid is unable to parse a PDF, it prints an error message to stdout, rather than raising an exception or returning an http error status code. Printing this error message to stdout breaks sidecar communication since we rely on stdout for data passing.
2. If a PDF could not be parse, it is not included in the References response from the sidecar. This adds Reference objects to the response even if we could not parse the PDF.


cguedes (approved):

hammer:
I'm still seeing brittle behavior after this fix...

PR #105, opened 2023-06-07
refactor: store file content in atom

### LINKED ISSUE #32: Improve TipTap editor content replacement
We need to add a delay (setTimeout) in order to prevent a weir behaviour in the TipTap editor when we try to change the content (i.e. toggle files).

We should investigate a way to fix this. Maybe this issue won't exist if we move the TipTap editor instance to the redux store.

_Originally posted by @sehyod in https://github.com/refstudio/refstudio/pull/30#discussion_r1204476574_

sehyod:
This PR fixes #32 

This splits the state in a more atomic way, with the following:
 - core primitive atoms storing simple data
 - core action atoms doing simple operations
 - exported combined atom holding the logic and calling the core atoms

This also adds an `activePane` state, that enables the app to open new file in the pane that has been used in last.


src/panels/MainPanel.tsx
```diff
@@ -80,38 +99,56 @@ export function MainPanelPane({
...
+    return <Spinner />;
+  }
+
+  if (loadableFileContent.state === 'hasError') {
+    return <strong>Error: {String(loadableFileContent.error)}</strong>;
   }
 
-  if (activeFile.fileExtension === 'xml') {
-    return <TextView file={activeFile} />;
+  if (loadableFileContent.data.type === 'XML') {
```
danvk, line 138: Add:

```ts
const {data} = loadableFileContent;
```

to simplify the code below.

src/panels/MainPanel.tsx
```diff
@@ -80,38 +99,56 @@ export function MainPanelPane({
...
+    );
   }
 
-  if (activeFile.fileExtension === 'pdf') {
-    return <PdfViewer file={activeFile} pdfViewerRef={pdfViewerRef} />;
+  if (loadableFileContent.data.type === 'PDF') {
+    return <PdfViewer file={loadableFileContent.data} pdfViewerRef={pdfViewerRef} />;
   }
 
   // Use TipTap editor by default!
```
danvk, line 152: Since the tiptap default is applied when the file is loaded now, you can switch this to an exhaustiveness check. The advantage of this is that if we add a new file type in `FileContent`, TypeScript will force us to add a new case here. Because of our eslint rules, this also pushes us to use a `switch` / `case` instead of `if` statements, which I think is fine (see https://github.com/typescript-eslint/typescript-eslint/issues/4682).

```ts
  const { data } = loadableFileContent;
  switch (data.type) {
    case 'XML':
      return <TextView file={data} />;
    case 'JSON':
      return <TextView file={data} textFormatter={(input) => JSON.stringify(JSON.parse(input), null, 2)} />;
    case 'PDF':
      return <PdfViewer file={data} pdfViewerRef={pdfViewerRef} />;
    case 'TIPTAP':
      return <TipTapView editorRef={editorRef} file={data} />;
    default:
      assertNever(data);
  }
```

Here's an implementation of `assertNever`:

```ts
/** Use this to assert that a branch is unreachable (for exhaustiveness checking). */
function assertNever(val: never) {
  throw new Error(`Reached the unreachable: ${val}`);
}
```

src/types/FileContent.ts
```diff
@@ -0,0 +1,21 @@
+export interface TipTapFileContent {
+  type: 'TIPTAP';
```
danvk, line 2: Can we use lowercase tags? This feels like yelling :)

src/views/PdfViewer.tsx
```diff
@@ -57,24 +52,12 @@ export function PdfViewer({ file, pdfViewerRef }: PdfViewerProps) {
     setNumPages(pdf.numPages);
   };
 
-  if (loadFileState.state === 'loading') {
```
danvk, line 60: 🎉  nice to see this logic unified!

src/atoms/core/paneGroupAtom.ts
```diff
@@ -0,0 +1,79 @@
...
+    id: paneId,
+    files: pane.openFiles.map((id) => fileEntries.get(id)).filter(isNonNullish),
+    activeFile: pane.activeFile ? fileEntries.get(pane.activeFile) : undefined,
+    activeFileAtom: pane.activeFile ? openFiles.get(pane.activeFile) : undefined,
+  };
+}
+
+export const _addFileToPane = atom(null, (get, set, { fileId, paneId }: PaneFileId) => {
+  const panes = get(_paneGroupAtom);
+  set(_paneGroupAtom, {
```
danvk, line 46: Suggest factoring out an `_updatePaneGroup` atom to DRY out the object spreading here, this should work:

```ts
const _updatePaneGroup = atom(null, (get, set, { paneId, ...update }: { paneId: PaneId } & Partial<PaneState>) => {
  const panes = get(_paneGroupAtom);
  set(_paneGroupAtom, {
    ...panes,
    [paneId]: {
      ...panes[paneId],
      ...update,
    },
  });
});
```

danvk (changes requested):
Nice to see the atoms wired up! Lots of suggestions but at a high level:

- Now that #100 is in, this really needs some tests. It would be great to have tests for some of the invariants that you're maintaining, for example that a file gets purged from memory once it's closed in both panes. OK if it's in another PR, but it should probably the very next PR!
- I think we should use an eslint rule to enforce that "core" atoms are only used in the expected places, rather than relying on comments and underscores. See comment below.
- I believe Jotai's `atomFamily` is designed to eliminate some of the bookkeeping around `Map`s.

It's less scary to refactor once you have tests, so maybe a path forward could be to add some unit tests for `src/atoms/fileActions.ts` (the public state API) and then refactor the internals to use `atomFamily`.

sehyod:
Thank you for these comments. I have addressed them
About the `atomFamily`, I decided not to use them because of this comment: https://github.com/pmndrs/jotai/discussions/1912#discussioncomment-5810504, I thought a map would be great to be able to iterate through open files. However, now the code is written, I realise we probably won't need it, as panes contain the list of open files and we can iterate through this list directly.
As you suggested, I think it would be better to refactor to use atomFamilies once we have tests. And I think this PR is already quite large so we should add tests in another PR. I have created issues for that: #128 and #129

cguedes (dismissed):

cguedes (approved):

PR #114, opened 2023-06-07
Generate reference citation keys for ingested references

### LINKED ISSUE #112: Generate reference citation key for ingested references
From #109 
> We could automatically generate a key on upload, and then allow the users to edit the key if they want. A common choice for a key is the first author's last name plus the publication year, e.g. hammerbacher1999. If there's a key collision just append a, b, c, etc. If we don't have metadata just use untitled plus 1, 2, 3, etc.

Looking at the 9 PDFs I have been using locally, Grobid only captures a date for 2 of them. I think these PDFs came from arxiv, so maybe that's part of the reason? Regardless, we'll need to improve this with the medium-term task of canonicalize references against some database.

gjreda:
fixes #112 

Adds `citation_key` and `published_date` to the Reference response.

Depending on whether the "uniqueness" postfix of `a, b, c ... 1, 2, 3` should be handled by the frontend or sidecar, this PR might require more work. See discussion [here](https://github.com/refstudio/refstudio/issues/112#issuecomment-1583051518).


.husky/pre-push
```diff
@@ -4,5 +4,4 @@
 yarn check
 yarn test
 
-yarn pylint:check
```
gjreda, line 7: Noticed that [someone added this](https://github.com/refstudio/refstudio/blob/main/package.json#L11) to `yarn check` (thank you!), so we don't need to call it twice here

python/sidecar/storage.py
```diff
@@ -25,6 +43,29 @@ def load(self):
...
+        """
+        for i, source in enumerate(self.references):
+            if source.filename_md5 == target.filename_md5:
+                logger.info(f"Updating reference {source.source_filename} ({source.filename_md5})")
+                self.references[i] = target
+                break
+        else:
+            logger.info(f"Unable to find reference {target.source_filename} ({target.filename_md5})")
+            return
+        self.save()
```
gjreda, line 68: @cguedes @sehyod for now, i'm assuming that you will be passing back in the entire `Reference` object, inclusive of `chunks` and such. it feels a little heavy, but seemed easiest for now. let me know if that's an issue for y'all.

python/sidecar/typing.py
```diff
@@ -17,19 +18,21 @@ class Reference(RefStudioModel):
...
 
 
 class Author(RefStudioModel):
     full_name: str
-    given_name: str
-    surname: str
-    email: str
+    given_name: str | None = None
+    surname: str | None = None
+    email: str | None = None
```
gjreda, line 35: Grobid won't always have these fields, but from what I can tell `full_name` will always be present if it can parse any authors.

We'll be able to get a lot more consistent here once we start work on further augmenting our references with #140

src/api/ingestion.ts
```diff
@@ -7,7 +7,7 @@ function parsePdfIngestionResponse(response: IngestResponse): ReferenceItem[] {
   return response.references.map((reference) => ({
     id: reference.filename_md5,
     title: reference.title ?? reference.source_filename.replace('.pdf', ''),
-    authors: (reference.authors ?? []).map((author) => ({ fullName: author.full_name, surname: author.surname })),
+    authors: (reference.authors ?? []).map((author) => ({ fullName: author.full_name })),
```
gjreda, line 10: @cguedes @sehyod I modified this to remove `surname` since it is not guaranteed we will have it even though we might have `reference.authors`. Feel free to push a commit if there's a better way to do this!

python/sidecar/storage.py
gjreda: alternatively, we could pass in the filename hash (or we should probably use a [UUID](https://docs.python.org/3/library/uuid.html)) and dict, making the function signature something like:
```python
def update(uuid: UUID, updates: dict[string, Any])
```
where `dict[string, Any]` => `field_name_to_update: value_to_set`. would need to think through how to do this for subtypes like Author though. maybe individual methods for `update_reference` and `update_author` to make things clear and specific.

cguedes: I think it is ok, also bc this is JsonStorage file. Maybe we need to expose to the frontend more granular operations (ex: rename citation). So this is ok for now.

cguedes (approved):

cguedes:
@gjreda just found an error (date parsing) in the ingestion for the [Machine Learning at Scale.pdf](https://github.com/refstudio/refstudio/files/11790287/Machine.Learning.at.Scale.pdf).

image

cguedes (changes requested):
There is a pending error when parsing some PDF files (check my latest comment).

cguedes (approved):

PR #127, opened 2023-06-09
feat: add JsonStorage and ability to chat with References

### LINKED ISSUE #82: Add ability to chat with your Reference PDFs [sidecar]
This issue is meant to capture the python implementation work to chat with your References.

Some related discussion:
- https://github.com/refstudio/refstudio/issues/61

### Chat: chat with your References

**Action**:
- User asks a question in the chat window
- Assumption: For now, we can assume the question is to all References. This might change in the future.

**Input**:
- text, question from the user (For now ... I could see the draft document being useful here).

**Sidecar Response**:
- json, single dict inclusive of the References used to answer

Example Response:
```json
{
  "answer": "I'm the answer to your question",
  "sources": [
    {
      "source_filename": "source1.pdf"
    },
    {
      "source_filename": "source2.pdf"
    }
  ]
}
```

### LINKED ISSUE #102: Backend data storage for reference content
As part of #82, we need some form of data storage for parsed reference content, particularly the chunked document text. Since #83 , we no longer store the chunked reference text. We'll need to add this back in, with or without embeddings.

We need to be able to query the chunked text in order to retrieve the most likely related chunks for a given chat question (regardless of whether we are using embeddings or something like bm25).

Some options and notes:
1. Structured text file on disk (e.g. JSON) - Because the sidecar is not a long running process, this file will need to be read from disk on each `chat` call. That doesn't feel great, though it might actually perform fine.
2. [LanceDB](https://github.com/lancedb/lancedb) - We used LanceDB in an [initial prototype](https://github.com/gjreda/scratch-pdf-bot) and it worked well. It was easy to get started, performed well, and had nice integration with Python data ecosystem. However, LanceDB currently does not have delete or update capabilities for individual records and is a fairly new project in very active development, so it is likely that its API will evolve quickly. It is focused on [vector](https://lancedb.github.io/lancedb/embedding/) and [full-text search](https://lancedb.github.io/lancedb/fts/), and does have a TypeScript client.
3. [Sqlite + VSS](https://github.com/asg017/sqlite-vss) - Can't say much about VSS as I'm not familiar with it, but Sqlite would be plenty sufficient for our needs.

gjreda:
fixes #102 and fixes #82 

Note that you will need to upload some PDFs and rerun PDF ingestion first. That will create a `references.json` which is needed for chat interactions.

There is more backend work to do related to chat, but I've captured that in other issues (see comment below). I didn't want this PR to continue to increase in scope and also think it'd be good to have this piece in so that we can begin to demo it.

```bash
❯ yarn python

❯ ./src-tauri/bin/python/main-aarch64-apple-darwin/main chat --text "What can you tell me about hidden feedback loops in machine learning?" | jq
[
  {
    "index": 0,
    "text": "Hidden feedback loops in machine learning refer to situations where two systems indirectly influence each other through the world, leading to changes in behavior that may not be immediately visible. These loops may exist between completely disjoint systems and can make analyzing the effect of proposed changes extremely difficult, adding cost to even simple improvements. It is recommended to look carefully for hidden feedback loops and remove them whenever feasible."
  }
]
```

- [x] Add tests for `chat.py`
- [x] Add tests for `ranker.py`
- [x] Add tests for `storage.py`
- [x] Write response to stdout
- [x] Add source filename or title to response (or create as new issue)
- [x] Create issue to capture idea of "named threads"
- [x] How does this handle "chat history?"
- [ ] How does this handle statements or non-questions?


gjreda:
In order to move this PR forward, I've created two follow-up issues for myself.

1. https://github.com/refstudio/refstudio/issues/143
2. https://github.com/refstudio/refstudio/issues/144

I think there's an open question on how and whether we want to handle chat interactions that are not questions, but IMO that might be something that is informed after spending some time with this implementation.

python/sidecar/ranker.py
```diff
@@ -0,0 +1,20 @@
...
+
+class BM25Ranker:
+    def __init__(self, storage: JsonStorage):
+        self.storage = storage
+        self.ranker = BM25Plus(self.storage.tokenized_corpus)
+    
+    def get_top_n(self, query: str, limit: int = 5) -> list:
+        """
+        Rank documents based on input text
+        :param input_text: str
```
sehyod, line 13: nit: the actual param name is query, not input_text

python/sidecar/storage.py
```diff
@@ -0,0 +1,33 @@
...
+class JsonStorage:
+    def __init__(self, filepath: str):
+        self.filepath = filepath
+        self.references = []
+        self.corpus = []
+        self.tokenized_corpus = []
+
+    def load(self):
+        with open(self.filepath, 'r') as f:
+            data = json.load(f)
```
sehyod, line 15: Should we have a try except here, in case the filepath or the file content is incorrect?

sehyod (approved):

cguedes (approved):
Everything looks good. My only comment is for the chat return type being an array instead of an object that contains an array.

PR #134, opened 2023-06-09
feat: make all editor blocks draggable

### LINKED ISSUE #133: The 6-gray dot block handle should only appear on hover.

sehyod:
Fixes #133 

Root document can now only contain draggable blocks. A draggable block is a wrapper for any block, thats adds a handle on the left:
- the handle only appear on hover
- clicking on the handle selects the block
- the handle always appear when the block is selected

This also adds rules when pressing enter:
- when the cursor is at the end of block, create an empty draggable block after
- when the cursor is in a block, split it in 2 draggable blocks
- if some content is selected, removes it, then split in 2 draggable blocks
These rules were what felt the most natural to me, but if you feel like anything seems weird when playing around with it, please let me know.

Note: This PR does not change the behaviour of blocks, which means that collapsible blocks are still hard to work with (no proper way to create a collapsible block, pressing Enter doesn't work, ...) and bullet points list are considered a whole block, which means you can only drag and drop the whole list. These components will be reworked in future PRs

image


cguedes:
@sehyod found this bug working this lists

http://g.recordit.co/g65Yjff0PD.gif

sehyod:
> @sehyod found this bug working this lists
> 
> g65Yjff0PD

Thanks for the catch! But to fix this, the best way would be to extend the List component to add custom behaviour, so I think it would be better to do it in another PR. That's actually the same with collapsible blocks: if you try to press Enter inside a collapsible block with the code of this PR, the result will be very buggy

src/TipTapEditor/DraggableBlock/DraggableBlockNode.ts
```diff
@@ -0,0 +1,112 @@
...
+            preNode = node.type.create(node.attrs, preNode, node.marks);
+          }
+
+          // Create a new draggable block, containing text after the selection
+          const postNode = this.type.create(
+            null,
+            $to.parent.type.create($to.parent.attrs, $to.parent.slice($to.parentOffset).content, $to.parent.marks),
+          );
+
+          if (dispatch) {
```
cguedes, line 80: Maybe you can push this check to the top.

```js
if(!dispatch) {
  return
}
```

cguedes (approved):

src/TipTapEditor/DraggableBlock/DraggableBlockNode.ts
```diff
@@ -0,0 +1,112 @@
...
+  parseHTML() {
+    return [
+      {
+        tag: 'div[data-type="draggable-block"]',
+      },
+    ];
+  },
+
+  renderHTML({ HTMLAttributes }) {
+    return ['div', mergeAttributes(HTMLAttributes, { 'data-type': 'draggable-block' }), 0];
```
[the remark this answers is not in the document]
sehyod: Fair point, it will be more consistent with the collapsible nodes

src/TipTapEditor/DraggableBlock/DraggableBlockNode.ts
sehyod: No, we actually want to return true even if dispatch is null or undefined. Prosemirror's commands can do dry-runs by passing a null dispatch. In this case, the command should still return the boolean indicating whether the command can run. From what I have seen from TipTap's repository, the usual way to do it it to have some checks first, that can make the command return false, then the transformations in an `if (dispatch) {` and finally `return true;`. Here is an example for the updateAttributes command: https://github.com/ueberdosis/tiptap/blob/b0198eb14b98db5ca691bd9bfe698ffaddbc4ded/packages/core/src/commands/updateAttributes.ts#L43

cguedes (approved):

PR #137, opened 2023-06-12
feat: 131 open editable files in the left and non editable ex pdfs in the right

### LINKED ISSUE #131: Open editable files in the left and non-editable  (ex PDFs) in the right
Originally

> Let's remove the full 2-pane layout for now and just allow PDFs to open into a separate read-only pane. We can revisit more complex layouts later when we're ready for more testing. Alternatively, we could consider adding a PDF reader as a 3rd element on the left sidebar, though that might be more complicated for now...

We already have the implementation support for "flexible left/right tab panes", allowing any file to be opened in any tab (including open the same file in both panes) and also split (i.e. move file to the other pane). 

So this issue will only restrict the usage in order to simplify the interaction with files in the editor.

cguedes:
Fixes #131 

With this PR PDF files are opened in the right panel, while the others open on the left. Note that the split file feature is still available as an action from the open files tree.

This PR also adds unit tests for the atom `fileActions` (and indirectly to the related core atoms). I've added also an action atom to close all files `closeAllFilesAtom`.

image

### DIFF
--- a/src/atoms/fileActions.test.ts
+++ b/src/atoms/fileActions.test.ts
@@ -234,9 +234,6 @@ describe('fileActions', () => {
   });
 });
 
-const PDF_FILE = makeFile('file.pdf');
-const TXT_FILE = makeFile('file.txt');
-
 function makeFile(name: string, extension?: string): FileFileEntry {
   name = extension ? name + '.' + extension : name;
   return {
--- a/src/atoms/fileActions.test.ts
+++ b/src/atoms/fileActions.test.ts
@@ -1,0 +1,41 @@
+import { act, renderHook } from '@testing-library/react';
+import { useAtomValue, useSetAtom } from 'jotai';
+import { describe, expect, test } from 'vitest';
+
+import { fileEntryAtom } from './core/fileEntry';
+import { openFileAtom, rightPaneAtom } from './fileActions';
+import { FileFileEntry } from './types/FileEntry';
+import { PaneId } from './types/PaneGroup';
+
+const pdfFile: FileFileEntry = {
+  name: 'file.pdf',
+  path: './file.pdf',
+  fileExtension: 'pdf',
+  isFolder: false,
+  isDotfile: false,
+  isFile: true,
+};
+describe('fileActions', () => {
+  test('should open PDF file as fileEntry', () => {
+    const { result } = renderHook(() => useSetAtom(openFileAtom));
+    const setOpenFile = result.current;
+
+    act(() => setOpenFile(pdfFile));
+
+    const { result: fileEntryResult } = renderHook(() => useAtomValue(fileEntryAtom));
+
+    expect(fileEntryResult.current.has(pdfFile.path)).toBe(true);
+    const fileEntry = fileEntryResult.current.get(pdfFile.path);
+    expect(fileEntry).toEqual({ ...pdfFile });
+  });
+
+  test('should open PDF file in the right pane', () => {
+    const { result: setOpenFileResult } = renderHook(() => useSetAtom(openFileAtom));
+    act(() => setOpenFileResult.current(pdfFile));
+
+    const { result: rightPanelResult } = renderHook(() => useAtomValue(rightPaneAtom));
+    expect(rightPanelResult.current.id).toBe(<PaneId>'RIGHT');
+    expect(rightPanelResult.current.files.length).toBe(1);
+    expect(rightPanelResult.current.files).toStrictEqual([pdfFile]);
+  });
+});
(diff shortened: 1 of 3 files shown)


hammer:
Thanks I think this reduction in feature complexity is a good one!

A thought while trying out the PR: I wonder if there's an easy way to add a way to close all reading tabs, similar to clicking on the icons on the left to collapse the left sidebar? I guess we're starting to get too many kinds of UI elements with a left sidebar, a second reading pane, and a right sidebar, and it would be good to unify or standardize them. No action to take now, just thinking out loud.

cguedes:
> A thought while trying out the PR: I wonder if there's an easy way to add a way to close all reading tabs, similar to clicking on the icons on the left to collapse the left sidebar? I guess we're starting to get too many kinds of UI elements with a left sidebar, a second reading pane, and a right sidebar, and it would be good to unify or standardize them. No action to take now, just thinking out loud.

We can have an action in the "OPEN FILES" panel header like vs code.

image

This is very easy to implement as we already have the operation `closeAllFilesAtom`. Should I create a task for this?

src/atoms/fileActions.test.ts
```diff
@@ -0,0 +1,329 @@
...
+    const { result: openFile } = renderHook(() => useSetAtom(openFileAtom, { store }));
+    const { result: leftPane } = renderHook(() => useAtomValue(leftPaneAtom, { store }));
+
+    const fileToOpen = makeFile('file.txt');
+    expect(leftPane.current.files.length).toBe(0);
+
+    act(() => {
+      openFile.current(fileToOpen);
+      openFile.current(fileToOpen);
+      openFile.current(fileToOpen);
```
sehyod, line 80: Wouldn't 2 calls be enough?

src/atoms/fileActions.test.ts
```diff
@@ -0,0 +1,329 @@
...
+
+  test('should not fail trying to close file that is not opened in pane', () => {
+    const store = createStore();
+    const { result: openFile } = renderHook(() => useSetAtom(openFileAtom, { store }));
+    const { result: closeFileFromPane } = renderHook(() => useSetAtom(closeFileFromPaneAtom, { store }));
+    const { result: leftPane } = renderHook(() => useAtomValue(leftPaneAtom, { store }));
+
+    const fileA = makeFile('fileA.txt');
+    const fileB = makeFile('fileB.txt');
+    const fileC = makeFile('fileC.txt');
```
sehyod, line 116: I think there is no need to create 3 files, this only makes the test have to check more things than it should

src/atoms/fileActions.test.ts
```diff
@@ -0,0 +1,329 @@
...
+      openFile.current(fileC);
+    });
+
+    expect(leftPane.current.activeFile).toEqual(fileC);
+
+    act(() => {
+      closeFileFromPane.current({ paneId: leftPane.current.id, fileId: fileC.path });
+    });
+
+    expect(leftPane.current.activeFile).not.toEqual(fileC);
```
sehyod, line 221: I think you can remove this line: you just need to check that the activeFile is fileB (which is done on the next line)

src/atoms/fileActions.test.ts
```diff
@@ -0,0 +1,329 @@
...
+  leftPaneAtom,
+  openFileAtom,
+  rightPaneAtom,
+  selectFileInPaneAtom,
+  splitFileToPaneAtom,
+} from './fileActions';
+import { FileFileEntry, FolderFileEntry } from './types/FileEntry';
+import { PaneId } from './types/PaneGroup';
+
+vi.mock('../filesystem');
```
sehyod, line 21: I have never worked with vitest so maybe this is irrelevant, but shouldn't we setup the mocks in a `beforeEach` or at least `beforeAll` and with something to clear mocks in `afterEach`. It seems very dangerous to me to have this `vi.mock` living here.

src/atoms/fileActions.test.ts
```diff
@@ -0,0 +1,329 @@
...
+    });
+
+    const fileA = makeFile('fileA.txt');
+
+    act(() => {
+      openFile.current(fileA);
+    });
+
+    expect(leftPane.current.activeFileContent).toBeDefined();
+    const { result: activeFile } = renderHook(() => useAtomValue(leftPane.current.activeFileContent!, { store }));
```
sehyod, line 296: The render syntax is a bit heavy to me. Maybe it would make sense to create utils function `renderSetAtomHook`, `renderGetAtomHook` and `renderAtomHook` that would take an atom and a store and return `result` directly. What do you think?

sehyod (changes requested):

hammer:
> Should I create a task for this?

Sure!

sehyod (approved):

Issue #142, opened 2023-06-12
Action to close all reading tabs


cguedes:
> A thought while trying out the PR: I wonder if there's an easy way to add a way to close all reading tabs, similar to clicking on the icons on the left to collapse the left sidebar? I guess we're starting to get too many kinds of UI elements with a left sidebar, a second reading pane, and a right sidebar, and it would be good to unify or standardize them. No action to take now, just thinking out loud.

_Originally posted by @hammer in https://github.com/refstudio/refstudio/issues/137#issuecomment-1587221741_

This can be implemented by adding an action similar to VS Code in the OPEN FILES header.

![image](https://github.com/refstudio/refstudio/assets/174127/d4d62575-ccf3-4e61-a314-7e4bb20739f1)

PR #147, opened 2023-06-13
feat: add button to close all files to Open Files panel

cguedes:
This PR closes issue #142 by adding a close all files action in the OPEN FILES panel.

image

### DIFF
--- a/src/components/PanelSection.tsx
+++ b/src/components/PanelSection.tsx
@@ -1,4 +1,5 @@
 import React, { useState } from 'react';
+import { IconType } from 'react-icons';
 import { VscChevronDown, VscChevronRight } from 'react-icons/vsc';
 
 import { cx } from '../cx';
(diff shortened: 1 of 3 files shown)


src/components/PanelSection.tsx
```diff
@@ -23,14 +26,29 @@ export function PanelSection({
     >
       <div
         className={cx(
+          'group/header',
```
sehyod, line 29: Is this from tailwind? Because you haven't defined any rule for the class

src/panels/ExplorerPanel.tsx
```diff
@@ -35,7 +37,10 @@ export function ExplorerPanel() {
 
   return (
     <PanelWrapper title="Explorer">
-      <PanelSection title="Open Files">
+      <PanelSection
+        rightIcons={[{ key: 'closeAll', Icon: VscCloseAll, title: 'Close All Open Files', onClick: closeAllFiles }]}
```
sehyod, line 41: I think the button should be removed when no file is open

src/components/PanelSection.tsx
cguedes: Yes, I also learned this one today :-). We need this bc there are two nested groups.

src/panels/ExplorerPanel.tsx
sehyod: Ok, fair enough!

sehyod (approved):

Issue #144, opened 2023-06-12
Capture chat history and create "named threads" for Chat interactions


gjreda:
Currently chat interactions are independent. We do not persist the message response from the LLM and thus we do not include potentially relevant past responses - a history - in future chat interations that are related.

1. 2. We'd like capture chat history, so that it can be used when it is relevant for future chat interactions.
2. We'd like to have a concept of "named threads" where a chat interaction creates a "conversation" with an autogenerated but relevant name (much like chat.openai.com).

PR #152, opened 2023-06-13
feat: add chat interaction UI

### LINKED ISSUE #148: Include chat in AI interaction component
With the chat backend merged (https://github.com/refstudio/refstudio/pull/127), we should expose this functionality in the app.

We reviewed various UX approaches for rewrite and chat at https://github.com/refstudio/refstudio/issues/78. I think the approach used by [Wordcraft](https://github.com/refstudio/refstudio/issues/78#issuecomment-1577403720) and [Type](https://github.com/refstudio/refstudio/issues/78#issuecomment-1577406795) of having AI interactions concentrated in the right sidebar is a good one.

For the initial chat UI, I can see us either doing what Type does and putting the selection into the chat box with the opportunity to add a chat interaction below it, or keeping rewrite and chat separate within the AI interaction pane.

If we go the Type direction, we should probably include some buttons for common rewrite actions within the box containing the selected text.

cguedes:
This PR closes #148 by adding a UI for chat interaction.

Note that currently the backend don't support threads and history (will be fixed by #144) so this PR adds that as a UI state in order to experiment the interaction pattern. 

image

### DIFF
--- a/src/api/rewrite.ts
+++ b/src/api/rewrite.ts
@@ -1,0 +1,10 @@
+import { callSidecar } from './sidecar';
+
+export async function askForRewrite(selection: string): Promise<string> {
+  try {
+    const response = await callSidecar('rewrite', ['--text', String(selection)]);
+    return response[0].text;
+  } catch (err) {
+    return `Rewrite error: ${String(err)}`;
+  }
+}
--- a/src/components/PrimarySideBar.test.tsx
+++ b/src/components/PrimarySideBar.test.tsx
@@ -2,26 +2,26 @@ import { render, screen, userEvent } from '../utils/test-utils';
 import { PrimarySideBar } from './PrimarySideBar';
 
 describe('PrimarySideBar', () => {
-  it('should display Explorer and References icons', () => {
+  test('should display Explorer and References icons', () => {
     render(<PrimarySideBar activePane="Explorer" onClick={() => 0} />);
     expect(screen.getByRole('menubar')).toBeInTheDocument();
     expect(screen.getByRole('menuitem', { name: 'Explorer' })).toBeInTheDocument();
     expect(screen.getByRole('menuitem', { name: 'References' })).toBeInTheDocument();
   });
 
-  it('should display active icon with 100% opacity ', () => {
+  test('should display active icon with 100% opacity ', () => {
     render(<PrimarySideBar activePane="References" onClick={() => 0} />);
     expect(screen.getByRole('menubar')).toBeInTheDocument();
     expect(screen.getByRole('menuitem', { name: 'References' })).toHaveClass('opacity-100');
   });
 
-  it('should display inactive icon with 50% opacity ', () => {
+  test('should display inactive icon with 50% opacity ', () => {
     render(<PrimarySideBar activePane="References" onClick={() => 0} />);
     expect(screen.getByRole('menubar')).toBeInTheDocument();
     expect(screen.getByRole('menuitem', { name: 'Explorer' })).toHaveClass('opacity-50');
   });
 
-  it('should call onClick with pane name', async () => {
+  test('should call onClick with pane name', async () => {
     const fn = vi.fn();
     render(<PrimarySideBar activePane="References" onClick={fn} />);
 
--- a/src/index.css
+++ b/src/index.css
@@ -12,6 +12,15 @@
   }
 }
 
+@layer components {
+  .btn-primary {
+    @apply cursor-pointer rounded-xl bg-primary hover:bg-primary-hover;
+  }
+  .btn-primary[disabled] {
+    @apply pointer-events-none opacity-50;
+  }
+}
+
 h1,
 h2,
 h3,
--- a/src/panels/ai/ChatPanelSection.test.tsx
+++ b/src/panels/ai/ChatPanelSection.test.tsx
@@ -44,6 +44,17 @@ describe('ChatPanelSection component', () => {
     expect(chatWithAiMock.mock.calls.length).toBe(1);
   });
 
+  // https://testing-library.com/docs/user-event/keyboard/
+  test('should display ERROR if chatWithAI throw exception', async () => {
+    const chatWithAiMock = vi.mocked(chatWithAI).mockRejectedValue('Error message.');
+    const user = userEvent.setup();
+    render(<ChatPanelSection />);
+    await user.type(screen.getByRole('textbox'), 'This is a question.');
+    await user.type(screen.getByRole('textbox'), '{enter}');
+    expect(chatWithAiMock.mock.calls.length).toBe(1);
+    expect(screen.getByText('ERROR: Error message.')).toBeInTheDocument();
+  });
+
   test('should NOT call api on SHIFT+ENTER in the textbox', async () => {
     const chatWithAiMock = vi.mocked(chatWithAI).mockResolvedValue([]);
     const user = userEvent.setup();
--- a/src/panels/ai/ChatPanelSection.test.tsx
+++ b/src/panels/ai/ChatPanelSection.test.tsx
@@ -1,0 +1,104 @@
+import { screen, waitFor } from '@testing-library/react';
+
+import { chatWithAI } from '../../api/chat';
+import { render, userEvent } from '../../utils/test-utils';
+import { ChatPanelSection } from './ChatPanelSection';
+
+vi.mock('../../api/chat');
+
+describe('ChatPanelSection component', () => {
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  test('should render the panel without history content and a textbox', () => {
+    render(<ChatPanelSection />);
+    expect(screen.getByText(/your chat history will be shown here\./i)).toBeInTheDocument();
+    expect(screen.getByRole('textbox')).toBeInTheDocument();
+  });
+
+  test('should call api if textbox has text', async () => {
+    const chatWithAiMock = vi.mocked(chatWithAI).mockResolvedValue([]);
+    const user = userEvent.setup();
+    render(<ChatPanelSection />);
+    await user.type(screen.getByRole('textbox'), 'This is a question.');
+    await user.click(screen.getByTitle('Send'));
+    expect(chatWithAiMock.mock.calls.length).toBe(1);
+  });
+
+  test('should NOT call api if textbox has empty text', async () => {
+    const chatWithAiMock = vi.mocked(chatWithAI).mockResolvedValue([]);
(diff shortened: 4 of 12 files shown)


src/api/rewrite.ts
```diff
@@ -0,0 +1,10 @@
+import { callSidecar } from './sidecar';
+
+export async function askForRewrite(selection: string): Promise<string> {
```
cguedes, line 3: Moved this function to the `/api` in order to have all the sidecar interaction in the api.

src/index.css
```diff
@@ -12,6 +12,15 @@
   }
 }
 
+@layer components {
+  .btn-primary {
```
cguedes, line 16: preliminar work to have custom tailwind components

tailwind.config.js
```diff
@@ -18,9 +18,23 @@ export default {
...
+  plugins: [debugPlugin(), autocompleteCustomComponentsPlugin()], 
 };
 
+/**
+ * Custom plugin that provide auto-complete support for the listed classes.
+ *
+ * The style definitions can be found in `index.css` under the `@layer components {` section
+ * 
+ */
+function autocompleteCustomComponentsPlugin() {
```
cguedes, line 30: With this you can now access the custom tailwind components in intellisense

image

src/panels/ai/ChatPanelSection.tsx
```diff
@@ -0,0 +1,121 @@
...
+  question: string;
+  answer?: string;
+}
+type ChatThread = ChatThreadItem[];
+
+export function ChatPanelSection() {
+  const [text, setText] = useState('');
+
+  const [chatThread, setChatThread] = useState<ChatThread>([
+    // {
```
cguedes, line 19: You can uncomment this to see the UI without having to use an api key.

sehyod (changes requested):
Thank you for this PR and for the work to add custom tailwind components!

src/panels/ai/ChatPanelSection.tsx
```diff
@@ -0,0 +1,121 @@
...
+    //     'I am an AI language model and do not have a name. How may I assist you with your query related to the given context?',
+    // },
+    // {
+    //   id: '2',
+    //   question: 'What can you tell me about hidden feedback loops in machine learning?',
+    //   answer:
+    //     'Hidden feedback loops in machine learning refer to situations where two systems indirectly influence each other through the world, leading to changes in behavior that may not be immediately visible. These loops may exist between completely disjoint systems and can make analyzing the effect of proposed changes extremely difficult, adding cost to even simple improvements. It is recommended to look carefully for hidden feedback loops and remove them whenever feasible.',
+    // },
+  ]);
+  const [currentChatThreadItem, setCurrentChatThreadItem] = useState<ChatThreadItem | undefined>();
```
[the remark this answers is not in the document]
cguedes: Ok. I can do that.

src/panels/ai/ChatPanelSection.tsx
```diff
@@ -0,0 +1,121 @@
...
+        'bg-gray-200': actor === 'ai',
+      })}
+    >
+      <div className="flex shrink-0 items-start self-start">
+        {actor === 'user' && <VscMegaphone size="24" />}
+        {actor === 'ai' && <VscWand size="24" />}
+      </div>
+      <div>
+        {text !== undefined && <div>{text}</div>}
+        {text === undefined && <div>...</div>}
```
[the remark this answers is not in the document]
cguedes: ok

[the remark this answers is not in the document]
cguedes: I rewrote to `<div>{text ? text : '...'}</div>`

src/panels/ai/ChatPanelSection.tsx
```diff
@@ -0,0 +1,120 @@
...
+    //     'I am an AI language model and do not have a name. How may I assist you with your query related to the given context?',
+    // },
+    // {
+    //   id: '2',
+    //   question: 'What can you tell me about hidden feedback loops in machine learning?',
+    //   answer:
+    //     'Hidden feedback loops in machine learning refer to situations where two systems indirectly influence each other through the world, leading to changes in behavior that may not be immediately visible. These loops may exist between completely disjoint systems and can make analyzing the effect of proposed changes extremely difficult, adding cost to even simple improvements. It is recommended to look carefully for hidden feedback loops and remove them whenever feasible.',
+    // },
+  ]);
+  const [currentChatThreadItem, setCurrentChatThreadItem] = useState<ChatThreadItem | null>();
```
sehyod, line 32: I think you have to pass `null` as the initial value, otherwise the type of currentChatThreadItem will be `ChatThreadItem | null | undefined`

hammer:
Very exciting! We're getting close to the MLP.

I hit an error when I tried to chat before uploading any references:
```
Chat error: Error: Traceback (most recent call last): File "main.py", line 13, in <module> File "sidecar/chat.py", line 30, in ask_question File "sidecar/storage.py", line 14, in load FileNotFoundError: [Errno 2] No such file or directory: '/Users/gium/Library/Application Support/com.tauri.dev/project-x/.storage/references.json' [50699] Failed to execute script 'main' due to unhandled exception!
```

I was able to get this working with a [PDF](https://github.com/refstudio/refstudio/files/11743714/Large.Language.Models.scientific.knowledge.and.factuality.pdf) I uploaded though, very cool!
Screenshot 2023-06-14 at 10 53 32 AM

cguedes:
That's the expected behaviour if you didn't uploaded PDFs. I forgot to mention that in the PR description.

sehyod (approved):

hammer:
> That's the expected behaviour if you didn't uploaded PDFs. 

Rather than the unfriendly error message can we say something more useful about needing to upload PDFs to chat? It's not clear to me that's even the right behavior. We should probably just mention to the user they can upload PDFs to improve the chat experience and pass through their chat to get a response.

cguedes:
That's a good point Jeff. In the frontend we can't distinguish between errors in the backend. You can have this one but also an API Rate limit, or invalid API Key, ..

I think that we should add (reply error codes) to the #144 and then improve the UI on top of that.

hammer:
> In the frontend we can't distinguish between errors in the backend. 

Ah, perhaps we should file a broader issue for better sharing of error state between the frontend and backend?

sehyod (approved):

src/api/rewrite.ts
```diff
@@ -0,0 +1,10 @@
...
+  try {
+    const response = await callSidecar('rewrite', ['--text', String(selection)]);
```
danvk, line 5: The `String()` wrapper here isn't necessary (if the TS types are to be believed!).

src/panels/ai/ChatPanelSection.tsx
```diff
@@ -0,0 +1,130 @@
...
+    //     'I am an AI language model and do not have a name. How may I assist you with your query related to the given context?',
+    // },
+    // {
+    //   id: '2',
+    //   question: 'What can you tell me about hidden feedback loops in machine learning?',
+    //   answer:
+    //     'Hidden feedback loops in machine learning refer to situations where two systems indirectly influence each other through the world, leading to changes in behavior that may not be immediately visible. These loops may exist between completely disjoint systems and can make analyzing the effect of proposed changes extremely difficult, adding cost to even simple improvements. It is recommended to look carefully for hidden feedback loops and remove them whenever feasible.',
+    // },
+  ]);
+  const [currentChatThreadItem, setCurrentChatThreadItem] = useState<ChatThreadItem | null>(null);
```
danvk, line 32: Do you want to move the thread state into atoms? Presumably the interactions will grow more elaborate in the future and we'll want it to persist even if you close the component. Alternatively the thread state could live in the backend.

src/panels/ai/ChatPanelSection.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  function handleKeyDown(evt: React.KeyboardEvent<HTMLTextAreaElement>): void {
```
danvk, line 34: Nit: more concisely written as:

```ts
const handleKeydown: React.KeyboardEventHandler = (evt) => {
```

src/panels/ai/ChatPanelSection.tsx
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  function handleChat(question: string) {
+    if (!question || currentChatThreadItem) {
+      return;
+    }
+    setCurrentChatThreadItem({ id: String(chatThread.length + 1), question: text });
+    setText('');
+    chatWithAI(question)
```
danvk, line 48: Suggest using `async` / `await`.

src/panels/ai/ChatPanelSection.tsx
```diff
@@ -0,0 +1,130 @@
...
+      .then((reply) => {
+        setChatThread([...chatThread, { id: String(chatThread.length + 1), question: text, answer: reply[0] }]);
```
danvk, line 50: There may be a stale closure issue here if you send two chats before the first one comes back. Safer to use the function form of `setState`:

```ts
setChatThread(prevChatThread => [...prevChatThread, { id: String(prevChatThread.length + 1), question: text, answer: reply[0] }]);
```

(also below)

src/panels/ai/ChatPanelSection.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const user = userEvent.setup();
+    render(<ChatPanelSection />);
+
+    await user.type(screen.getByRole('textbox'), 'Will this test pass?');
+    await user.click(screen.getByTitle('Send'));
+
+    expect(screen.getByText('Will this test pass?')).toBeInTheDocument();
+    expect(screen.getByText('...')).toBeInTheDocument();
+
+    resolveFn();
```
danvk, line 109: Nice test!

src/components/PrimarySideBar.test.tsx
```diff
@@ -2,26 +2,26 @@ import { render, screen, userEvent } from '../utils/test-utils';
 import { PrimarySideBar } from './PrimarySideBar';
 
 describe('PrimarySideBar', () => {
-  it('should display Explorer and References icons', () => {
+  test('should display Explorer and References icons', () => {
```
danvk, line 5: I think I left a comment about this elsewhere, but when we write "should" in a test description we should use `it` rather than `test` to make it read more clearly (I pretty much always use `it`) https://stackoverflow.com/questions/45778192/what-is-the-difference-between-it-and-test-in-jest

danvk (commented):
A few more comments. I got this to work in my local client.

Issue #145, opened 2023-06-12
Move References Storage Path into an environment variable or settings


gjreda:
The path in which we store `references.json` is currently hardcoded because we do not currently have shared settings between the front-end and backend. See #108.

gjreda:
fixed by #153

Issue #108, opened 2023-06-07
Create settings configuration that can be shared by front-end and backend


gjreda:
With #104 we're starting to see the need for some settings to be shared by the tauri, the front-end, and backend sidecar. We should figure out a way of sharing this via some settings configuration file.

gjreda:
@hammer I think it's ok to close this one now since we ended up landing on `.env` files

PR #153, opened 2023-06-13
feat: configuration screen

cguedes:
This PR adds a configuration and underlying settings support to the app.

Features:
  - Settings core
    - Typed settings for (general, openAI and sidecar.logging)
    - Values read from .env with fallback to default value (first initialisation)
    - Settings saved in a JSON file
  - Settings UI
    - Opened (and closed) via menu shortcut `refstudio -> settings` or keyboard `Cmd+,` shortcut
    - Sidebar icon ⚙️ to open modal
    - Modal component that display settings in panel sections
    - Editable OpenAI and Sidecar logging values
  - Sidecar integration
    - settings are sent to sidecar via _environment variables_
      - `APP_DATA_DIR` - (default to Tauri's `appDataDir()`)
      - `PROJECT_NAME` - (default to project-x)
      - `OPENAI_API_KEY` - (default from `OPENAI_API_KEY` env)
      - `OPENAI_CHAT_MODEL` - (default from `OPENAI_CHAT_MODEL` or `gpt-3.5-turbo`)
      - `OPENAI_COMPLETE_MODEL` - (default from `OPENAI_COMPLETE_MODEL` env or `davinci`) 
      - `SIDECAR_ENABLE_LOGGING` - (default from `SIDECAR_ENABLE_LOGGING` env or `false`)
      - `SIDECAR_LOG_DIR` - (default from `SIDECAR_LOG_DIR` env or `/tmp`)
  - Rust
    - module for application menu (skeleton for file and edit actions)
    - handler for application menu actions and send then as events to the frontend via events integration
    - read env vars using dotenv
  - Utilities
    - `useAsyncEffect` utility to simplify async effects
    - `noop` to simplify tests with mandatory handler we don't need to use/test
  - Tests & coverage
    - 100% coverage 😅     

image

----

### Tasks

- [x] Config settings screen window
- [x] Global trigger with `Cmd+,`
- [x] Little gear icon down in the lower left corner like in VS Code
- [x] Config settings schema + settings API
- [x] settings storage
- [x] Config panel for OpenAI settings: `OPENAI_API_KEY`, `OPENAI_COMPLETE_MODEL`, `OPENAI_CHAT_MODEL`
- [x] Default settings read from env vars
- [x] Flow settings (as ENV) when calling sidecar
- [x] Register tauri menu action for **Settings**
- [x] Unit tests


hammer:
Nice! In addition to access to this screen from the keyboard, we should probably put a little gear icon down in the lower left corner like in VS Code.

Currently I configure 3 things with OpenAI based on what I saw in @gjreda's checkin: `OPENAI_API_KEY`, `OPENAI_COMPLETE_MODEL`, `OPENAI_CHAT_MODEL`. We should probably grow the list of OpenAI configuration parameters to include all 3.

Also, I configure these settings now through environment variables. Should this configuration screen auto-populate with what it finds in environment variables? If so, we should probably also add some text for each configuration variable to indicate what environment variable name we use. I tend to think of the configuration hierarchy as starting with environment variables, which can be overridden by config files, which can be overridden by command-line flags. Given that we expect most users to install this application and run it from a launcher or by clicking an icon, I'm not sure if it makes sense to give every configuration variable a corresponding environment variable, but it sure is convenient to not have to set these variables each time I try out the app...

I'm thinking at least for OpenAI we do honor environment variables for now.

cguedes:
I think that using environment variables as defaults (show in the config screen) is a nice to have (mostly for development - and maybe for some reasonable defaults if we include some sort of .env in the build). 

Adding command-line arguments seems a bit to much for now.

Also, I'm going to save the settings in a permanent storage (.json file) to persist settings over app runs.

src/components/PrimarySideBar.test.tsx
```diff
@@ -1,32 +1,40 @@
 import { render, screen, userEvent } from '../utils/test-utils';
 import { PrimarySideBar } from './PrimarySideBar';
 
+const emptyFn = vi.fn();
```
danvk, line 4: `vi.fn` is for mocking. Since you don't care about this function, suggest using `_.noop` from lodash instead.

src/components/PrimarySideBar.tsx
```diff
@@ -1,27 +1,29 @@
 import { IconType } from 'react-icons';
-import { VscFiles, VscLibrary } from 'react-icons/vsc';
+import { VscFiles, VscLibrary, VscSettingsGear } from 'react-icons/vsc';
 
 import { cx } from '../cx';
 
-export type PrimarySideBarPane = 'Explorer' | 'References';
+export type PrimarySideBarPane = 'Explorer' | 'References' | '';
```
danvk, line 6: What does empty string mean? Can you use `null` instead?

src/settings/SettingsModal.tsx
```diff
@@ -0,0 +1,254 @@
...
+    SETTINGS_PANES.flatMap((e) => e.panes).find((p) => p.id === 'project-openai')!,
+  );
+
+  useEffect(() => {
+    (async function runAsync() {
+      try {
+        await unregister(SETTINGS_SHORTCUT_TOGGLE);
+        await unregister(SETTINGS_SHORTCUT_CLOSE);
+        await register(SETTINGS_SHORTCUT_TOGGLE, () => onToggle(!open));
+        await register(SETTINGS_SHORTCUT_CLOSE, () => onToggle(false));
```
danvk, line 83: Presumably you don't want <kbd>ESC</kbd> to be registered globally, only when the Settings modal is open?

My suggestion would be:

- `SettingsModal` is always open when it's in the virtual DOM.
- Its visibility is owned by the parent (`App`) which also registers the <kbd>Cmd</kbd><kbd>,</kbd> shortcut (and presumably never needs to unregister it).
- `SettingsModal` registers the <kbd>ESC</kbd> shortcut and unregisters it when it unmounts. It can just have an `onClose` prop.

src/App.tsx
```diff
@@ -20,23 +21,37 @@ function App() {
...
+        <LeftSidePanelWrapper
+          onRefClicked={(reference) => editorRef.current?.insertReference(reference)}
+          onSettingsClick={() => setSettingsOpen(true)}
+        />
+        <Panel defaultSize={60}>
+          <MainPanel editorRef={editorRef} pdfViewerRef={pdfViewerRef} />
+        </Panel>
+        <RightPanelWrapper />
+      </PanelGroup>
+      <SettingsModal open={settingsOpen} onToggle={(open) => setSettingsOpen(open)} />
```
danvk, line 43: See comment in `SettingsModal`, I'd suggest making this:

```
{open && <SettingsModal ... />}
```

src/settings/settings.ts
```diff
@@ -0,0 +1,52 @@
...
+      name: 'project-x',
+    },
+    openAI: {
+      apiKey: await readEnv('OPENAI_API_KEY'),
+      completeModel: await readEnv('OPENAI_COMPLETE_MODEL'),
+      chatModel: await readEnv('OPENAI_CHAT_MODEL'),
+    },
+  },
+  {
+    fileName: 'settings',
```
danvk, line 27: Does this persist settings in `settings.json`?

danvk (dismissed):
I'm also curious about the relationship between `.env` and `settings.json`. Is the latter populated from the former only once when you first start the app? That may be unavoidable, but I can definitely imagine it being a source of confusion in the future.

If we anticipate having lots of settings, it might be convenient to use a library like this one to generate a basic form UI from the schema for the settings:
https://github.com/rjsf-team/react-jsonschema-form

I believe VS Code does something like this. Here's an example of how the settings are specified for an extension:
https://github.com/asciidoctor/asciidoctor-vscode/blob/50515506f5c55eb14caee419bd15bda67b501700/package.json#L267

src/components/PrimarySideBar.test.tsx
cguedes: You have a point, but I didn't wanteed to add a new dependency to the project (even though most likely we will need lodash in the future!)

src/settings/SettingsModal.tsx
cguedes: My (initial) idea was to have a drop-in component for the settings that handles the open/close behaviour. Meanwhile, with the sidebar action to open the settings I've have to move the open/close control to the App.

With the addition of the window menu entries and handles I rewrite this component to:
 - Stop using Tauri's global shortcut (that feature is for scenarios when you want a OS-wide shortcut trigger for the app)
 - Adopt Tauri's [events](https://tauri.app/v1/guides/features/events) for the communication between Rust and the front-end
 - New event for the settings menu click (tauri://menu/settings)
 - Emit settings menu event when on clicks in the Settings icon in the sidebar

src/components/PrimarySideBar.tsx
cguedes: Sure. This was a legacy implementation.

src/settings/settings.ts
cguedes: Yes, the lib appends the extension and ignores the extension if we define it.
I've also made the `dir` explicit to help discover where the file is.

cguedes:
> I'm also curious about the relationship between `.env` and `settings.json`. Is the latter populated from the former only once when you first start the app? That may be unavoidable, but I can definitely imagine it being a source of confusion in the future.

I see your confusion. The idea is to be able to bundle a .env file with default configurations that is read (once) to bootstrap de settings (settings.json). From that point onwards the app only read that file.

I think that for now, too keep it simple, we should have the defaults (ex: `OPENAI_COMPLETE_MODEL` and `OPENAI_CHAT_MODEL`) in the code and stop using the `.env` file. We can revisit this topic in the future when building config levels (i.e: user, app, project).

Also bc for the backend development (CC @gjreda) that file would be useful to store keys that would be received via sidebar when the app is running.

cguedes:
> If we anticipate having lots of settings, it might be convenient to use a library like this one to generate a basic form UI from the schema for the settings: https://github.com/rjsf-team/react-jsonschema-form
> 
> I believe VS Code does something like this. Here's an example of how the settings are specified for an extension: https://github.com/asciidoctor/asciidoctor-vscode/blob/50515506f5c55eb14caee419bd15bda67b501700/package.json#L267

I believe that using react-jsonschema-form is too much for now. We should keep this PR simpler.

hammer:
> I'm also curious about the relationship between .env and settings.json

In my experience, configuration is done in 3 places, with increasing precedence: environment variables, settings files, and command line options.

> we should have the defaults (ex: OPENAI_COMPLETE_MODEL and OPENAI_CHAT_MODEL) in the code and stop using the .env file.

We should not use a `.env` file ourselves but we should look in the environment variables to see if any are set.

Could we also give our settings file name something specific to the app?

gjreda:
> Sure, I can renamed the file to be something like refstudio-settings.json, but note that this file will be stored inside the [app's config dir](https://tauri.app/v1/api/js/path/#configdir).

@cguedes is there an easy way for us to share the appdir location between the front-end and backend (#108 and #145). Something that keeps coming up for the backend is that I don't have a way of knowing where appdir is, as I assume it will be different based on the OS.

For now, on `ingest`, I just chop off the `project-x` and use the remaining path for storing things, but later on in non-ingest things I might want to also write to the filesystem (e.g. chat history). It'd be great if there were a way I can fetch some `appdir` variable and use that for everything.

Another alternative would be to pass the `appdir` path from front-end to the backend when calls are made, but that feels not great.

cguedes:
> @cguedes is there an easy way for us to share the appdir location between the front-end and backend (https://github.com/refstudio/refstudio/issues/108 and https://github.com/refstudio/refstudio/issues/145). Something that keeps coming up for the backend is that I don't have a way of knowing where appdir is, as I assume it will be different based on the OS.

I'm planning to send `APP_DATA_DIR` (ex: `~/Library/Application Support/com.tauri.dev` in OSX) and `PROJECT_NAME` (ex: `project-x`) to you via env variables in the sidecar call from the frontend.

I can send additional directory paths as you need them.

gjreda:
I'm probably not the best person to review the front-end code here, but the way we are doing the new `env` passing works for me.

src/components/PrimarySideBar.test.tsx
```diff
@@ -1,32 +1,42 @@
 import { render, screen, userEvent } from '../utils/test-utils';
 import { PrimarySideBar } from './PrimarySideBar';
 
+const noop = () => {
+  // Empty
+};
+
```
sehyod, line 7: You can use the `noop` function from the utils file

sehyod (approved):

src/components/PrimarySideBar.test.tsx
cguedes: You are right! Fixed.

sehyod (dismissed):

sehyod (approved):

Issue #151, opened 2023-06-13
Esc should toggle between block browsing and text editing


hammer:
When using Notion, hitting `Esc` will highlight the current block and move into a "block browsing" mode where operations on the tree of blocks that form the document can be taken. Hitting `Esc` when in this mode will return the user back to text editing.

I don't think this keyboard shortcut is high priority but just filing now as I'm thinking of it.

PR #154, opened 2023-06-13
feat: improve collapsible blocks

### LINKED ISSUE #139: Typing ">" + space should create a new collapsible element or turn an existing block into a collapsible element.

sehyod:
Fixes #139 

- Typing `> ` now creates a collapsible (or transforms the current block into a collapsible block)
- The arrow's colour becomes lighter when the content is empty
- Pressing Backspace at the beginning of a collapsible block removes the block

This is a WIP branch, the behaviours when pressing Enter and when deleting content across nodes containing a collapsible block still need to be defined.

### DIFF
--- a/src/TipTapEditor/CollapsibleBlock/CollapsibleBlock.css
+++ b/src/TipTapEditor/CollapsibleBlock/CollapsibleBlock.css
@@ -17,6 +17,10 @@ collapsible-block > button.folded:before {
   content: '\25b6';
 }
 
+collapsible-block:not(:has(collapsible-content)) > button:before {
+  color: #93928e;
+}
+
 collapsible-block .content {
   @apply flex-1;
 }
--- a/src/TipTapEditor/CollapsibleBlock/CollapsibleBlock.tsx
+++ b/src/TipTapEditor/CollapsibleBlock/CollapsibleBlock.tsx
@@ -1,7 +1,8 @@
 import './CollapsibleBlock.css';
 
 import { Node } from '@tiptap/pm/model';
-import { NodeViewContent, NodeViewWrapper } from '@tiptap/react';
+import { TextSelection } from '@tiptap/pm/state';
+import { Editor, NodeViewContent, NodeViewWrapper } from '@tiptap/react';
 import { ElementType } from 'react';
 
 import { cx } from '../../cx';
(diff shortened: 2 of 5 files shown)


hammer:
Nice! My Notion-like editor experience dreams are becoming a reality.

Could you say more about the situation of deleting content across nodes containing a collapsible block? I'm not sure I understand the situation well enough to comment. When in doubt, try it in Notion!

A few differences I notice from Notion:
- In Notion, if my cursor is at the end of the text of a collapsible section and I hit `cmd+enter`, it opens the collapsible section, and if I hit `cmd+enter` again, it toggles it back to collapsed. If I hit `enter`, I get a new, empty collapsible section:

https://github.com/refstudio/refstudio/assets/15233/fe968a0a-c162-4c7d-9ff7-1c53610b1131
- In Notion, if I hit `tab`, the item indents, and if I hit `shift+tab`, it un-indents.
- In Notion, `Esc` highlights the block and allows the keyboard to be used to move the block up and down: https://github.com/refstudio/refstudio/issues/151

I also hit what I believe to be a bug: I have an unordered list inside of a collapsible element. When the cursor is at the end of the collapsible element text, I hit `enter`. Instead of bumping the list down and allowing me to edit directly above it, it just deletes the unordered list.

https://github.com/refstudio/refstudio/assets/15233/7aeca920-338a-4e50-81e3-a077ca7e738c

sehyod:
Thank you for the feedback!

What I meant for the situation of deleting across nodes was this situation:

https://github.com/refstudio/refstudio/assets/58954208/b57d8fe9-9ad6-4571-8343-ed668d962263

What happens in notion is that the content of the last collapsible block is added as the content of the collapsible block where the start of the selection is.

But this actually brings another question: what should happen when the start of the selection is not in collapsible block?

This is the result on notion and it seems a bit weird, as you can see the empty collapsible block is moved above 1.1.3 for some reason:

https://github.com/refstudio/refstudio/assets/58954208/b2e9bde3-4d7a-4be5-bd06-ea2ef67aea50

- For the cmd+enter, that should be pretty easy to add! Thank you for letting me know the shortcut :)
- We currently don't have any indentation concept: collapsible blocks are blocks with a header and multiple draggable blocks as content. I think we still need to get more comfortable with prosemirror's api to understand what is achievable or not
- Thanks for opening the issue about node browsing!
- About the bug, it's because the behaviour of pressing `Enter` has not been implemented in this PR yet, so the results is quite weird. From what I have seen on notion, pressing enter when the node is uncollapsed should add a new empty section, as you mentioned, and pressing `Enter` on a collapsed node, should add a new collapsible node below. Is that correct?

hammer:
> This is the result on notion and it seems a bit weird

That is weird! I don't have a great thought for the right behavior in this situation. The "ghost" collapsible caret with the text dangling below it doesn't seem ideal but maybe there are deeper reasons Notion does that. I think we're fine if we write down the decision we make and file it away to revisit later.

> From what I have seen on notion, pressing enter when the node is uncollapsed should add a new empty section, as you mentioned, and pressing Enter on a collapsed node, should add a new collapsible node below. Is that correct?

Sounds correct to me.

danvk:
Just played around with this a bit. The copy/paste behavior with folds is nice!

A few remaining differences I noticed from Notion, probably not all related to this PR:

- Typing should hide the drag handle until you move the mouse again
- When you create a fold by typing <kbd>></kbd> at the start of a line, it should be open by default.

- When you type <kbd>></kbd> at the start of a bulleted list item, it shifts down to the next line rather than turning the list item into a fold as Notion does:

https://github.com/refstudio/refstudio/assets/98301/edfd3f0f-9cef-4d64-ace1-36ee3c17c8d8

- When you hit <kbd>Enter</kbd> at the end or the middle of a list item, it should add a new list item, rather than an item outside the list.

https://github.com/refstudio/refstudio/assets/98301/137e1016-ba42-4986-9c96-8c797c6817e4

- Backspacing over the "•" at the start of a bulleted list item should remove that item the bulleted list (or move it onto the previous item in the bulleted list if it's not empty):

https://github.com/refstudio/refstudio/assets/98301/5240a72b-7690-4250-a215-5d678c22d871

- In Notion, the content under a fold isn't indented. Only the fold caret itself is:
image

vs. in RefStudio the content _and_ the caret are indented:
image

- When you select text, there are broken rectangles for lines that aren't completely selected (the blue under "Ite" on the first line):
image

Notion just selects the text, rather than the full width of the editor for each line. This looks neater.

- Typing `## heading<enter>` leaves you in H2 style in RefStudio whereas it drops you back into paragraph style in Notion (which is the better behavior).

- Selecting three lines of text and clicking the "Bulleted List" button in the toolbar should make a bulleted list with three items (it currently does nothing).

danvk (commented):
This is definitely a strong move in the direction of Notion parity!

The code for manipulating the nodes is quite intricate and my main question is around how we want to set up testing for it. A few ideas:

- Have a "script" that a human periodically goes through ("type ABC, put the cursor between B and C and hit enter"). This would be tedious but would at least document our expectations. This is kind of what @hammer and I are doing already in the comments :)
- Look into some kind of test runner like [Playwright](https://playwright.dev/). These sorts of things are always a bit of a PITA to set up but they do create an extremely realistic test environment. I believe this is what VS Code uses. TipTap looks like it [uses Cypress](https://github.com/ueberdosis/tiptap/blob/develop/tests/cypress/integration/core/extendMarkRange.spec.ts), which is similar.
- Try to create some sort of unit test by creating the nodes that tip tap expects (maybe by writing out Markdown) and calling the individual functions like `setCollapsibleBlock` with each situation (no selection, selected text on the fold header line, selection crossing the header line and foldable content). Depending on how TipTap works internally, this may or may not be possible

hammer:
I pulled @danvk's comments on Notion parity into a new issue: https://github.com/refstudio/refstudio/pull/154#issuecomment-1593557137. Feel free to address the relevant ones here and do the rest in a future PR!

cguedes:
Collapsible block content is indented in RefStudio. Notion doesn't do that and I agree that we should'n also.

image

Add'l, I feel that the spacing between blocks (e.g collapsible child) is a bit to big.

src/editor/nodes/collapsibleBlock/inputRuleHandlers/chevronHandler.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    expect(editor.state.selection.empty).toBe(true);
+    const { $from } = editor.state.selection;
+
+    // The selection should point to the start of the text
+    expect(editor.state.doc.nodeAt($from.pos)?.type.name).toEqual('text');
+    expect($from.sameParent(editor.state.doc.resolve($from.pos - 1))).toBe(false);
+  });
+
+  test('should do nothing when the block cannot be turned into a collapsible block', () => {
```
cguedes, line 34: why can't we turn headings into collapsible? @sehyod

hammer:
We have follow up issues in case it’s preferable to get this PR in and address some of the other issues later https://github.com/refstudio/refstudio/issues/168 and https://github.com/refstudio/refstudio/issues/169

src/editor/nodes/collapsibleBlock/inputRuleHandlers/chevronHandler.test.ts
sehyod: This is because headers are blocks, whereas collapsible summary must be inline content. I have checked on notion and this is aligned with what they do (you cannot turn a header into a collapsible block)

cguedes: That is only true for H1 (the page title) in Notion.

cguedes:
You are right @hammer. We will move these comments to the other issue(s) and merge this (it's actually merged already 🎉 ).

Issue #136, opened 2023-06-10
Find replacement for dataclasses-jsonschema


hammer:
https://github.com/refstudio/refstudio/pull/126 added a dependency on https://github.com/s-knibbs/dataclasses-jsonschema, which is in maintenance mode. @gjreda already hit an issue where the new `T | None` syntax for optional types is not handled by `dataclasses-jsonschema`: https://github.com/refstudio/refstudio/pull/135.

While I like the idea of pulling Python types into JavaScript, I think it's a nice to have and not a must have at this stage of the project, and I'm nervous about hitting more issues due to this dependency.

I don't want to spend much time on it but if we could either find a replacement for this dependency or remove it, I think that would allow us to stay focused on features on the critical path for the project right now.

gjreda:
We should be able to switch to Pydantic pretty easily and it will generate jsonschema for us: https://docs.pydantic.dev/latest/usage/schema/

PR #161, opened 2023-06-15
Clean up pydantic JSON Schema and generated TypeScript types

danvk:
We were getting some odd type aliases on the TypeScript side because Pydantic generated JSON Schema with fields like this:

```json
{
  "given_name": {
    "title": "Given Name",
    "type": "string"
  }
}
```

json-schema-to-typescript turns this into:

```ts
type GivenName = string;
```

and then uses that as a type alias. The party line from both projects is that they're both doing this correctly (see https://github.com/bcherny/json-schema-to-typescript/issues/504 and https://github.com/pydantic/pydantic/issues/1740), however pydantic seems to recognize that this may not be what everyone wants and provides example code to strip off `title` in their docs:

"For example, the title key can be removed from the model's properties:"
https://docs.pydantic.dev/latest/usage/schema/#schema-customization

We can define our own base model to use this in all our API types.


gjreda (approved):

gjreda:
part of this work for #136

PR #164, opened 2023-06-15
usePromise, useCallablePromise → react-query

danvk:
Rather than use our own bespoke Promise hooks, we should probably just use react-query, which is wildly popular and supports this use case just fine:
https://stackoverflow.com/questions/62340697/react-query-how-to-usequery-when-button-is-clicked

There's a slight behavior change here: previously if you fetched a rewrite and then selected some new text, we'd show the new text and the old rewrite. Now we clear the rewrite. If we want to go back to the old behavior, we could change to `queryKey: []` or `queryKey: [!!selection]`.


src/panels/ai/SelectionPanelSection.tsx
```diff
@@ -19,18 +31,14 @@ export function SelectionPanelSection() {
...
-            disabled={rewriteCallState.state === 'loading'}
-            onClick={() => callRewrite(debouncedSelection)}
-          >
+          <button className="btn-primary" disabled={isFetching} onClick={fetchRewrite}>
             REWRITE
           </button>
-          {rewriteCallState.state === 'error' && <span className="bg-red-50">{String(rewriteCallState.error)}</span>}
-          {rewriteCallState.state === 'loading' && <span>Processing...</span>}
-          {rewriteCallState.state === 'ok' && (
+          {error ? <span className="bg-red-50">{String(error)}</span> : null}
```
sehyod, line 37: Why not write `!!error && <span className="bg-red-50">{String(error)}</span>` here?

cguedes: Updated.

cguedes (approved):

Issue #113, opened 2023-06-07
Ability to display Reference ingestion status


gjreda:
From #109 

> The ingestion pipeline should display more complete information about its status while running, and we need to display information about how references have been processed after the ingest pipeline has finished, e.g. the PDF is available in the local project directory, metadata was parsed successfully from GROBID, content has been extracted and made available for AI interactions, etc.

hammer:
I probably shouldn't have combined those two into a bullet, as they are pretty different in implementation. The first request is about references that are being processed, and the second request is about the reference list once processing is done.

gjreda:
Some thoughts on the "states" a Reference can be in. Naming conventions are off the top of my head, so feel free to suggest others!

1. `staged_for_ingest` - Uploaded and ready for Grobid parsing, but Grobid parsing not yet attempted.
2. `grobid_processed` - Grobid parsing was attempted and ran successfully. This state does not aim to tell us specifics about what metadata was extracted.
3. `grobid_failure` - Grobid parsing was attempted and failed. Grobid does not tell us much about why a PDF failed, so there's not much we can say here.
4. `available_for_chat` - Reference content has been extracted and will be used as part of `chat`.

In the future, I imagine that we'll have additional statuses, but I don't think we should solve for these now as we do not yet have ingest pieces for them.

gjreda:
Dropping notes from Slack thread with @cguedes and @sehyod

1. Update IngestResponse to include a status for each Reference - that the current Reference payload at the end of ingest should also include a status
2. Add a new endpoint to retrieve all References that have previously been processed (these are just stored as json on the filesystem, so y'all might just be able to fetch them as well).
3. Add a new endpoint to do a status check on References during ingestion, with the intention that this would be called while the full ingestion pipeline is running, so that the UI can update as necessary.

> generally yes. The idea is the be able to initiate an ingestion process (1) and get status for the background processing (3) by calling a different endpoint (bc the first call needs to keep running for the ingestion). The (2) is an extra safeguard we can run when the app starts but not really needed for now IMO.

> For (1) I think that you won’t be able to return the references information before the processing finish so we can implement (2) to get all references with each ref containing the current processing status. And (3) to check if there is any pending action in progress with a lightweight and specialised payload.

PR #178, opened 2023-06-20
feat: Add ingestion status for References

gjreda:
This is still a WIP for now. Need to fix one bug and add/update tests.
_____
part of #113

This PR updates ingest to:
1. Copy `uploads` and work from a `.staging` directory so that each new upload does not process previously ingested References
2. Adds a `status` field to `Reference` which displays the "completed" ingestion status
3. Be a bit cleaner and easier to read


python/sidecar/ingest.py
```diff
@@ -26,90 +32,210 @@
...
+        jstore.load()
+        return jstore.references
 
-    def get_files_to_ingest(self) -> list[Path]:
+    def _get_files_to_ingest(self) -> list[Path]:
         """
-        Determines which files need to be ingested
-        :return: bool
+        Determines which files need to be ingested by comparing the list of 
+        already processed References against files found in `uploads`.
```
cguedes, line 137: Wouldn't be easier to give you the list of file paths to ingest? In the FE the user selects files and not directories.

python/cli.schema.json
```diff
@@ -82,6 +82,9 @@
         "citation_key": {
           "type": "string"
         },
+        "status": {
+          "type": "string"
```
cguedes, line 86: Can we have a enum here (i.e. expected strings for status)?

python/sidecar/ingest.py
```diff
@@ -26,90 +32,210 @@
 handler.setFormatter(formatter)
 
 logger.addHandler(handler)
-logger.disabled = os.environ.get("SIDECAR_ENABLE_LOGGING", "false").lower() == "true"
+logger.disabled = os.environ.get("SIDECAR_DISABLE_LOGGING", "true").lower() == "true"
```
cguedes, line 35: Do you want to make logging active by default?

gjreda:
Since we merged #181 for fetching ingest status, I am going to close this PR and break out the pieces of it that are related to #218 has a new PR.

Issue #173, opened 2023-06-20
Improve upload target for references


cguedes:
> Upload target: as the library grows, the upload target will move down the screen. Should we consider moving to a tabbed layout with "upload" as a tab next to "library"? Or some other solution? Also, I'd like to make the whole component a target for files that are dragged over it.

From https://github.com/refstudio/refstudio/issues/109

cguedes:
Ok.

For 1, and given the solution proposed for 2 (#118), I think that we should just remove the upload section (not the feature!) and implement something like this (still not the final version).


a) Initial screen with a "welcome message" and information about how to upload (click or drag & drop).

<img width="300" alt="image" src="https://github.com/refstudio/refstudio/assets/174127/e050c5db-4385-4e08-8469-0f665eff0fa8">

b) upload placeholder for guidance on drag/drop

<img width="1156" alt="image" src="https://github.com/refstudio/refstudio/assets/174127/c293ae3b-3259-433f-ba39-31db27fdb300">

c) Right after upload, instead of just displaying a spinner, we show a _cleaner_ progress bar and the files selected (with some placeholder data). We can improve this #113 

<img width="300" alt="image" src="https://github.com/refstudio/refstudio/assets/174127/3d2b5e37-52d7-4da4-a5d6-d04a38dd09a1">

d) After first upload, only the guidance message is shown in the bottom. We should try to ingest ~100 of PDFs to "stress test the UI". We can maybe add a "References > Upload" menu action to trigger the file upload flow.

<img width="300" alt="image" src="https://github.com/refstudio/refstudio/assets/174127/323772dd-46d3-4a10-b7de-7438e6db0576">

cguedes:
Reopen as we have found a regression in the TipTap editor drag-and-drop.

cguedes:
Dropping a note about the upcoming Tauri feature in v2.0 that would provide the file paths on hover during the drag-and-drop.

https://github.com/tauri-apps/tauri/pull/7601/files#diff-3d3c40eb99396b09a2d8f63bcb9c41b6a9895e1b4af1e0ac833d8c3d1f63e9dfR1457-R1461

![image](https://github.com/refstudio/refstudio/assets/174127/c83584c2-dc7f-4a8d-a672-9428722b94b4)

Issue #159, opened 2023-06-15
References table view and reference detail view


hammer:
As part of #109 we need to revisit the references list to be a table view. We also need to consider a reference detail view where we display all known metadata about a reference and allow the user to edit metadata. Metadata includes both scholarly metadata as well as application metadata such as whether we have the PDF, whether GROBID is parsing or was able to parse metadata, whether the PDF contents are available for chat, etc.

Let's collect screenshots and sketches of the potential UX for each of these components here.

At a high level we discussed whether the references table view should be in the left sidebar where it currently resides or if it might be better handled by a "terminal view" that comes up from the bottom.

For the reference detail view, we discussed either having a collapsible section drop down below a reference in the table when it is clicked or opening up an editor window to view details for an individual reference.

---
Tasks (added by @cguedes) after meeting

- [x] "List view" of many references in left sidebar: list view with sort, filter, add, export functionality at the top (#202)
- [x] "Table view" of many references in editor area: full table with add/remove columns, sort/filter/tag/organize, bulk modifications, etc. (#200, #201)
- [x] "Detail view" of one reference in editor area: edit metadata for a single reference. In the future: add notes, full text, tags, etc. ( #182,  #199)

hammer:
Here's how Zotero does it. They have a simple table layout with very few default columns. New columns can be added by right-clicking on the column names and selecting more. Sorting can be done by clicking a column, and a secondary sort column can be set by right-clicking a column title.

There are two layout options for the reference detail view: the standard view puts the details to the right of the table, and the stacked view puts them to the bottom. Within the detail view, there's a standard form with fields to edit, and a set of tabs for notes, tags, and related references. Note that related references are manually added by the user.

<img width="1470" alt="Screenshot 2023-06-16 at 1 39 19 PM" src="https://github.com/refstudio/refstudio/assets/15233/c3054cda-d9a9-40ee-ae06-218cf94a5cfd">
<img width="1470" alt="Screenshot 2023-06-16 at 1 39 08 PM" src="https://github.com/refstudio/refstudio/assets/15233/ee53af3e-69c8-4efa-bbda-0fd11b2cd098">

hammer:
### Papers

[...]

Issue #118, opened 2023-06-08
Drag and drop PDF takes over editor with no way out


danvk:
Drag and Drop a PDF onto the "RS" logo or the "Explorer" file view (I think anywhere other than the sanctioned "upload" box in the references panel).

The window turns into a PDF viewer and there's no way to get back to the RefStudio app other than closing it and re-opening.

We should either `preventDefault` on all drag-and-drop interactions in the Tauri window or allow PDFs to be dropped anywhere to import them.

hammer:
My preference is to allow PDFs to be dropped anywhere. I've seen other apps handle this situation when a PDF is dragged over app window by graying or blurring the app and displaying a message like "Add PDF to references" or something like that.

PR #179, opened 2023-06-21
feat: improve references upload flow

cguedes:
Note: This is still a WIP.


src/components/Spinner/index.tsx
```diff
@@ -1,5 +1,38 @@
...
   return <div className="lds-dual-ring" />;
 }
+
+export function ProgressSpinner({ badge }: { badge: string }) {
+  const STEP = 1;
+  const [percent, setValue] = useState(0);
+
+  useInterval(() => {
+    setValue(Math.min(percent + STEP, 100));
+  }, 200);
```
sehyod, line 16: I know this is only a temporary solution, but I don't feel super confident about this potentially infinite interval. Could you do something like this instead?
```typescript
useInterval(() => {
    setValue(Math.min(percent + STEP, 100));
  }, percent < 100 ? 200 : null);
```
So that the interval stops when `percent` reaches 100

src/panels/references/ReferencesDropZone.tsx
```diff
@@ -0,0 +1,130 @@
...
+          if (existingRef) {
+            return null;
+          }
+          return {
+            id: filename,
+            citationKey: '...',
+            title: filename,
+            filename,
+            authors: [{ fullName: 'Unknown' }],
+          } as ReferenceItem;
```
sehyod, line 42: We shouldn't have to use `as ReferenceItem`. If references without a published date are valid items, we should make the field optional in the type definition.

src/panels/references/ReferencesDropZone.tsx
```diff
@@ -0,0 +1,130 @@
...
+  });
+
+  const copyAndIngestMutation = useMutation({
+    mutationFn: async (filePaths: string[]) => {
+      await copyFilesMutation.mutateAsync(filePaths);
+      await ingestMutation.mutateAsync();
+    },
+  });
+
+  // FIXME: Replace with use async effect (after settings PR merge)
```
sehyod, line 69: I think the PR has been merged so you can use useAsyncEffect now :)

src/panels/references/ReferencesDropZone.tsx
```diff
@@ -0,0 +1,130 @@
...
+          }
+        }),
+      ];
+    })();
+
+    return () => {
+      isMounted = false;
+      unlisten.forEach((fn) => fn());
+    };
+  }, [copyAndIngestMutation, visible, files, setSyncInProgress]);
```
sehyod, line 106: You must include `ingestMutation` in the dependencies list

sehyod (changes requested):

src/components/Spinner/index.tsx
cguedes: Yes, is temporary. But will fix that.

src/panels/references/ReferencesDropZone.tsx
cguedes: Did it already. And looks nice!

hammer:
Nice! This PR definitely improves https://github.com/refstudio/refstudio/issues/118. Eventually we might want to do some kind of file type detection as I am able to drag any kind of file onto the app and get the upload behavior which is probably not the best user experience.

This PR does not address the first part of https://github.com/refstudio/refstudio/issues/173:

> as the library grows, the upload target will move down the screen. Should we consider moving to a tabbed layout with "upload" as a tab next to "library"? Or some other solution?

In this PR, the text `TIP: Click here or drag/drop PDF files here for upload.` appears below the list of references and hence falls off the screen when there are many references. Perhaps when we do the references table view in https://github.com/refstudio/refstudio/issues/159 we can revisit how the system file browser is accessed and make it always visible?

For the ingestion status bar, if it's hard to communicate percent complete, I'm fine with just show that we're busy someplace. I guess we're going to have to figure out how to handle user actions that are taken while we are busing with the ingestion process. We certainly don't want to block the whole app, but perhaps we want to block new uploads until the previous uploads finish?

src/panels/references/ReferencesDropZone.tsx
cguedes: I've used as mostly for autocomplete of field. But you are right. Maybe we should make all these fields optional as they are optional in the source.

cguedes:
> Nice! This PR definitely improves https://github.com/refstudio/refstudio/issues/118. Eventually we might want to do some kind of file type detection as I am able to drag any kind of file onto the app and get the upload behavior which is probably not the best user experience.

Sure. That's fixed.

> In this PR, the text TIP: Click here or drag/drop PDF files here for upload. appears below the list of references and hence falls off the screen when there are many references. 

My idea, for the MVP, was that they would see this information when the references list is empty. So they will learn how to upload (click or drag-drop). I agree that we should add a menu entry for `References -> Upload` and also some other indicator (icon) in the panel for the upload action.

> For the ingestion status bar, if it's hard to communicate percent complete, I'm fine with just show that we're busy someplace. 

Yes, the idea was to show that we are busy. The %age isn't even working as we don't have that information (we will have after #178). 

> I guess we're going to have to figure out how to handle user actions that are taken while we are busing with the ingestion process. We certainly don't want to block the whole app, but perhaps we want to block new uploads until the previous uploads finish?

Sure. I decided that for now we won't block them from upload/ingest concurrently but def improve this after that backend part is improved. Also, for the "busy indicator" maybe we should build a fixed bottom bar like VS Code has for loading indicators.

image

sehyod (approved):

PR #180, opened 2023-06-21
feat: add reference viewer

### LINKED ISSUE #174: Improve click behaviour on references
> What should happen when we **click on a reference**? The two most likely actions would be to open the PDF for reading or to expand the reference to display more metadata about it. Maybe we do both, with icons?

From #109

sehyod:
Fixes #174


hammer:
Nice! This provides an early implementation of the "reference detail view" mentioned in https://github.com/refstudio/refstudio/issues/159. When we tackle the "references table view" mentioned in that issue we may want to revisit the click target to be an icon or button rather than the whole reference row, but that's a small detail we can target later. Ultimately we'll want this reference detail view to support editing the metadata as well of course, and we may want to allow the authors to be clicked to search for references by that author, etc.

sehyod:
Yes, you're right! This PR is the first step in having other things than files in the editor. We can definitely (and we probably will) add buttons to open either the pdf itself or the reference details.
For the next features, like being able to click on an author name, I think we need to implement the references table view first, because we could then open this view with a filtering set to the author name

cguedes (approved):

Issue #175, opened 2023-06-20
Extract raw PDF text even if Grobid parsing fails


gjreda:
We do not currently attempt to extract PDF text if Grobid parsing is unsuccessful. We can still do this via something like `pdfminer`, which would allow us to make the Reference content available for chat.

hammer:
Linking back to an old issue where we listed some options for PDF parsing for my own convenience: https://github.com/refstudio/refstudio/issues/14

gjreda:
Using `pypdf` for this as simple text extraction is much easier with it than it is with `pdfplumber`. `pypdf` is also what's used under the hood in both LangChain and LlamaIndex.

`pdfplumber` seems better suited if we wanted to go deep in the weeds on pdf parsing and roll our own logic, rather than simply extracting the text out. See [this discussion](https://github.com/jsvine/pdfplumber/issues/242) as an example.

PR #181, opened 2023-06-21
feat: Add new endpoint for ingestion status

### LINKED ISSUE #113: Ability to display Reference ingestion status

gjreda:
fixes #113 

Example response below. If an unhandled exception is raised when calling this endpoint, we'll return `"status": "error"` with an empty list of statuses.

```bash
❯ poetry run python main.py ingest --status | jq
{
  "status": "ok",
  "reference_statuses": [
    {
      "source_filename": "A Few Useful Things to Know about Machine Learning.pdf",
      "status": "complete"
    },
    {
      "source_filename": "2301.10140.pdf",
      "status": "pending"
    },
    {
      "source_filename": "grobid-fails.pdf",
      "status": "failure"
    }
  ]
}
```


python/sidecar/ingest.py
```diff
@@ -203,6 +222,7 @@ def _create_references_for_grobid_failures(self) -> list[Reference]:
             references.append(
                 Reference(
                     source_filename=source_pdf,
+                    status=typing.IngestStatus.FAILURE,
```
gjreda, line 225: I'm not convinced we should call this an `failure`. Just because Grobid was unable to process it does not mean we cannot do things with it. See [this issue for example](https://github.com/refstudio/refstudio/issues/175).

scripts/codegen.sh
```diff
@@ -5,6 +5,7 @@ set -x
 
 poetry run python python/generate_schema.py > python/cli.schema.json
 yarn run json2ts python/cli.schema.json --no-additionalProperties > src/api/types.ts
+yarn prettier --config package.json --write src/api/types.ts
```
gjreda, line 8: Fixes weird pre-push loop between codegen and prettier: https://ref-studio.slack.com/archives/C05525DV9HN/p1687469412858809?thread_ts=1687468688.106519&cid=C05525DV9HN

python/sidecar/ingest.py
cguedes: I agree that for now "failure" is the most appropriate since we are only doing Grobid.

src/api/types.ts
```diff
@@ -20,6 +25,7 @@ export interface IngestResponse {
 export interface Reference {
   source_filename: string;
   filename_md5: string;
+  status?: IngestStatus & string;
```
cguedes, line 28: Do you know why we have the extra `& string` in this definition?
Can we make this field mandatory? (I know we might need to reset the references file for this to work fine, but I think that it worth the trouble). Also, I saw above that the field has a default PROCESSING value.

python/main.py
```diff
@@ -1,14 +1,23 @@
...
     parser = cli.get_arg_parser()
     args = parser.parse_args()
+
     if args.debug:
         print(args)
+    
     if args.command == "ingest":
+        if args.status:
+            ingest.get_statuses()
+            sys.exit()
```
cguedes, line 15: I don't think that implementing this a an option of ingest would work with the CliCommands typings we are using. We can only vary the output type by command and not by parameters.

Also, you need to update the README.md to add this new command.

cguedes (changes requested):
* Update README
* Create a new command to be able to set the correct return type (CliCommands are per command)
* Try to make status field mandatory

src/api/types.ts
cguedes: Great. Thanks.

cguedes (approved):

PR #185, opened 2023-06-22
Add vitest and arrow eslint plugins

danvk:
This adds two new eslint plugins:

- [eslint-plugin-vitest](https://yarnpkg.com/package/eslint-plugin-vitest)
- [eslint-plugin-prefer-arrow](https://yarnpkg.com/package/eslint-plugin-prefer-arrow)

This was prompted by discovering two tests with the same name in #186.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -47,7 +47,7 @@ function LeftSidePanelWrapper() {
   const [primaryPane, setPrimaryPane] = useState<PrimarySideBarPane>('Explorer');
   const openReference = useSetAtom(openReferenceAtom);
 
-  function handleSideBarClick(selectedPane: PrimarySideBarPane) {
+  const handleSideBarClick = (selectedPane: PrimarySideBarPane) => {
     if (selectedPane === primaryPane) {
       // Toggle collapsing
       setPrimaryPaneCollapsed(!primaryPaneCollapsed);
@@ -56,7 +56,7 @@ function LeftSidePanelWrapper() {
       setPrimaryPaneCollapsed(false);
     }
     setPrimaryPane(selectedPane);
-  }
+  };
 
   React.useEffect(() => {
     if (primaryPaneCollapsed) {
--- a/src/api/chat.test.ts
+++ b/src/api/chat.test.ts
@@ -10,7 +10,7 @@ describe('chatWithAI', () => {
     vi.clearAllMocks();
   });
 
-  test('should return list of strings with content returned by sidecar', async () => {
+  it('should return list of strings with content returned by sidecar', async () => {
     vi.mocked(callSidecar).mockResolvedValue([
       {
         index: 0,
@@ -22,7 +22,7 @@ describe('chatWithAI', () => {
     expect(response).toStrictEqual(['some reply']);
   });
 
-  test('should return list of strings with content returned by sidecar', async () => {
+  it('should return list of strings with content returned by sidecar', async () => {
     vi.mocked(callSidecar).mockResolvedValue([
       {
         index: 0,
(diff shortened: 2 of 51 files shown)


src/api/chat.test.ts
```diff
@@ -22,7 +22,7 @@ describe('chatWithAI', () => {
     expect(response).toStrictEqual(['some reply']);
   });
 
-  test('should return list of strings with content returned by sidecar', async () => {
+  it('should return list of strings with content returned by sidecar 2', async () => {
```
danvk, line 25: This test had the same name as the previous one but different content.

src/App.tsx
```diff
@@ -47,7 +47,7 @@ function LeftSidePanelWrapper() {
   const [primaryPane, setPrimaryPane] = useState<PrimarySideBarPane>('Explorer');
   const openReference = useSetAtom(openReferenceAtom);
 
-  function handleSideBarClick(selectedPane: PrimarySideBarPane) {
```
danvk, line 50: function statements inside other functions are frowned upon because they're hoisted, just like variables declared with `var`.

src/api/chat.test.ts
```diff
@@ -10,7 +10,7 @@ describe('chatWithAI', () => {
     vi.clearAllMocks();
   });
 
-  test('should return list of strings with content returned by sidecar', async () => {
+  it('should return list of strings with content returned by sidecar', async () => {
```
danvk, line 13: The consistent-test-it rule says to use `it` inside `describe` blocks and `test` outside.

src/editor/nodes/collapsibleBlock/inputRuleHandlers/chevronHandler.test.ts
```diff
@@ -19,19 +19,19 @@ describe('Chevron input rule handler', () => {
     expect(collapsibleBlocks).toHaveLength(1);
 
     const [collapsibleBlock] = collapsibleBlocks;
-    expect(getText(collapsibleBlock)).toEqual('Some content');
+    expect(getText(collapsibleBlock)).toBe('Some content');
```
danvk, line 22: See https://github.com/veritem/eslint-plugin-vitest/blob/HEAD/docs/rules/prefer-to-be.md

src/editor/nodes/collapsibleBlock/nodes/CollapsibleBlockContent.ts
```diff
@@ -7,15 +7,11 @@ export const CollapsibleBlockContentNode = Node.create({
...
   selectable: false,
 
-  parseHTML() {
-    return [
-      {
-        tag: 'collapsible-content',
-      },
-    ];
-  },
+  parseHTML: () => [
```
danvk, line 10: I don't mind the left-hand side of this diff, but using arrow functions does allow you to drop the `return` for the common case of a one-liner.

src/editor/nodes/collapsibleBlock/keyboardShortcutCommands/enter.test.ts
```diff
@@ -50,38 +50,16 @@ describe('Enter keyboard shortcut command', () => {
...
     expect(content.childCount).toBe(3);
-    expect(getText(content.child(0))).toEqual('der');
-    expect(getText(content.child(1))).toEqual('Content Line 1');
-    expect(getText(content.child(2))).toEqual('Content Line 2');
+    expect(getText(content.child(0))).toBe('der');
+    expect(getText(content.child(1))).toBe('Content Line 1');
+    expect(getText(content.child(2))).toBe('Content Line 2');
   });
 
-  test('should add a new content line to an uncollapsed collapsible block', () => {
```
danvk, line 62: This was an exact duplicate of the previous test.

src/App.tsx
cguedes: What is wrong with that? Is there any other advantages of using function expression? (I know that is easier to define the function).

src/api/chat.test.ts
cguedes: That's ok. Thanks for the bulk rename :-).

cguedes (approved):

sehyod (approved):

PR #186, opened 2023-06-22
Rework some editor tests to use visual cursor and snapshots

### LINKED ISSUE #169: Testing node manipulation
From @danvk at https://github.com/refstudio/refstudio/pull/154#pullrequestreview-1482196425:

> The code for manipulating the nodes is quite intricate and my main question is around how we want to set up testing for it. A few ideas:
> 
> - Have a "script" that a human periodically goes through ("type ABC, put the cursor between B and C and hit enter"). This would be tedious but would at least document our expectations. This is kind of what @hammer and I are doing already in the comments :)
> - Look into some kind of test runner like [Playwright](https://playwright.dev/). These sorts of things are always a bit of a PITA to set up but they do create an extremely realistic test environment. I believe this is what VS Code uses. TipTap looks like it [uses Cypress](https://github.com/ueberdosis/tiptap/blob/develop/tests/cypress/integration/core/extendMarkRange.spec.ts), which is similar.
> - Try to create some sort of unit test by creating the nodes that tip tap expects (maybe by writing out Markdown) and calling the individual functions like setCollapsibleBlock with each situation (no selection, selected text on the fold header line, selection crossing the header line and foldable content). Depending on how TipTap works internally, this may or may not be possible

danvk:
This reworks some of the new editor tests from #154 to be more legible by making them more "visual". So rather than:

```ts
editor
  .chain()	
  .setContent(defaultParagraph + defaultParagraph)	
  // 18 is the beginning of the second paragraph	
  .setTextSelection(18)	
  .run();
```

We have:

```ts
setUpEditorWithSelection(editor, `<p>Some content</p><p>|Some content</p>`);
```

The `|` here indicates where the cursor should be. With two `|` characters you get a selection. This is somewhat similar in spirit to how TypeScript tests editor interactions (google "typescript fourslash").

This has a few advantages:

- More legible than a magic number
- Editing the editor content in a test doesn't require updating the selection positions by hand

On the assertion side, I set up snapshot testing. This tends to make it much clearer what the resulting state of the document after an interaction is than a series of assertions.

Closes #169


src/editor/test-utils.ts
```diff
@@ -38,5 +40,67 @@ export function findNodesByNodeType(node: Node, nodeType: string) {
...
+ * @returns The text positions of the one or two "|" characters in the editor.
+ */
+export function setUpEditorWithSelection(editor: Editor, contentHTML: string) {
+  editor.chain().setContent(contentHTML).run();
+  const minPos = TextSelection.atStart(editor.state.doc).from;
+  const maxPos = TextSelection.atEnd(editor.state.doc).to;
+
+  const positions = [];
+  for (let i = minPos; i <= maxPos; i++) {
+    const text = editor.view.state.doc.textBetween(i, i + 1);
```
danvk, line 63: This probably isn't the most efficient way to do this, but it works!

src/editor/nodes/collapsibleBlock/keyboardShortcutCommands/enter.test.ts
```diff
@@ -11,52 +11,63 @@ describe('Enter keyboard shortcut command', () => {
...
-    expect(collapsibleBlock).toBeDefined();
-    expect(collapsibleBlock.childCount).toBe(2);
-    expect(getText(collapsibleBlock.child(0))).toEqual('Hea');
-
-    const content = collapsibleBlock.child(1);
-    expect(content.childCount).toBe(3);
-    expect(getText(content.child(0))).toEqual('der');
-    expect(getText(content.child(1))).toEqual('Content Line 1');
-    expect(getText(content.child(2))).toEqual('Content Line 2');
+    expect(getPrettyHTML(editor)).toMatchInlineSnapshot(`
```
danvk, line 61: This is an example of a test that reads much more clearly after this change.

src/editor/nodes/collapsibleBlock/keyboardShortcutCommands/enter.test.ts
```diff
@@ -11,52 +11,63 @@ describe('Enter keyboard shortcut command', () => {
...
+      "<collapsible-block folded=\\"false\\">
```
cguedes, line 62: @danvk  @sehyod wondering if we can fix/improve these ' vs '' in the snapshot (would read better without the `\\"`) and also, would be great to match the `setUpEditorWithSelection`.

cguedes (approved):

src/editor/nodes/collapsibleBlock/keyboardShortcutCommands/enter.test.ts
sehyod: Yes, I 100% agree. I have pushed a change to use single quotes in `getPrettyHtml`

cguedes (approved):

PR #189, opened 2023-06-23
chore: add chakra ui

### LINKED ISSUE #3: Component libraries
Tailwind vs. Styled Components.
Dan used Material UI, which is closer to Styled Components, but made everything slow and big.
The YLD team likes [Chakra](https://chakra-ui.com/).

sehyod:
This adds Chakra UI
This also refactors the File Tree component to use Chakra UI and get to a more VSCode-like UI

https://github.com/refstudio/refstudio/assets/58954208/4182436d-48fa-4c73-bdff-79023994f4ed

Closes #3

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,9 @@
     "tauri:dev:debug": "DEV_TOOLS=1 tauri dev"
   },
   "dependencies": {
+    "@chakra-ui/react": "^2.7.1",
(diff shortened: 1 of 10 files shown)


sehyod:
With this PoC PR, we think that we should not switch to Chakra UI and keep using tailwind, at least for now

PR #192, opened 2023-06-23
fix: Fix regression issue with drag-drop in editor

cguedes:
The upload files feature broke the drag-drop of draggable elements in the TipTap editor!

* We've had to deactivate Tauri's fileDropEnabled 
* Implement drag-and-drop with custom HTML events (only for PDF files)
* Hover effect when drag-drop for user feedback
* This PR adds initial UI for the application footer with ingestion status
* This PR also adds a new app menu entry `References -> Upload...` to open the upload files dialog

https://github.com/refstudio/refstudio/assets/174127/7ed4afd3-6392-4cfb-b980-7700868aa168

NOTE: This PR don't aim to improve the left references sidebar.

This will close https://github.com/refstudio/refstudio/issues/173


sehyod (approved):
Looks good! I let two comments, but nothing blocking.
While testing locally, I realised that you can still drag and drop a file on the title bar and it opens it with no way of getting out. It's very specific, I don't think it has to be fixed in this PR, but we should create a ticket for it to fix it later on.

sehyod (approved):

PR #193, opened 2023-06-23
[Sidecar] Add ability to remove one or all references

### LINKED ISSUE #191: [Sidecar] Add ability to remove one or all references
Add sidecar endpoint to enable #156

gjreda:
fixes #191 

___
```bash
# Example:
$ poetry run python main.py delete --source_filenames grobid-fails.pdf "Machine Learning at Scale.pdf" | jq

# Response:
{
  "status": "ok",
  "message": ""
}

# Another Example (error):
$ poetry run python main.py delete --source_filenames file-does-not-exist.pdf | jq

# Error Response:
{
  "status": "error",
  "message": "Unable to delete file-does-not-exist.pdf: not found in storage"
}
```

### DIFF
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -19,7 +19,7 @@ export interface CliCommands {
   ingest_status: IngestStatusResponse;
   rewrite: RewriteChoice[];
   chat: ChatResponseChoice[];
-  delete: ResponseStatus;
+  delete: DeleteStatusResponse;
 }
 export interface IngestResponse {
   project_name: string;
@@ -67,4 +67,8 @@ export interface RewriteChoice {
 export interface ChatResponseChoice {
   index: number;
   text: string;
+}
+export interface DeleteStatusResponse {
+  status: ResponseStatus;
+  message: string;
 }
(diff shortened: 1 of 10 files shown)


sehyod (dismissed):

src/api/types.ts
```diff
@@ -19,6 +19,7 @@ export interface CliCommands {
   ingest_status: IngestStatusResponse;
   rewrite: RewriteChoice[];
   chat: ChatResponseChoice[];
+  delete: ResponseStatus;
```
cguedes, line 22: This is just returning a status and not what you have described in the README.md (status/message).

cguedes (changes requested):
Fix the CliCommands response payload.

Also, I would prefer to have an extra argument (ex: --all) to remove all files and another argument for selected files (--source_filesnames <filename> <filename> <filename> ...).

cguedes (approved):

cguedes (changes requested):
The current implementation only updates the references metadata (references.json) and don't remove any other files (pdf, .grobid, ...). 

This behaviour (i.e. the PDF inside uploads) makes the ingestion a new PDF (i.e. a different PDF) to also ingest previously deleted PDFs.

gjreda:
I think it makes sense for the front-end to delete the actual PDF from the filesystem, since the front-end is currently what writes the PDF to the `uploads` directory.

I agree that in the future this should probably all be owned by the backend though, particularly in a world where the backend is some set of web APIs.

cguedes (approved):

PR #204, opened 2023-06-27
feat: add ability to save files

### LINKED ISSUE #197: Add ability to save files
From #157 

Update in-memory files so that switching between tabs does not reset the content.
Add `Save` action to save "TipTap" files

sehyod:
This adds the ability to save files:
 - Files are saved in memory so that switching between tabs does not reset the content
 - Pressing cmd+S (or ctrl+S in non-macOS environment) saves the file on the disk
 - Saving is also available in the top menu, in `File`->`Save`

https://github.com/refstudio/refstudio/assets/58954208/46c0cab0-0689-4222-8eec-ecfdb87b2a92

---

Implementation choices:
Four atoms are available for files:
 - `loadableFileAtom` is the atom containing the loadable file (the state is `loading` when initially reading the file from disk, and should stay `hasData` afterwards)
 - `updateFileBufferAtom` updates a buffer atom. This is currently updated on every update in the editor. The need of a buffer comes from the fact that updating the actual atom triggers a re-render of the editor, that deletes the history and resets focus
 - `saveFileInMemoryAtom` saves the content of the buffer in the `loadableFileAtom`. This atom is currently called when an editor instance is unmounted (ie. when switching between tabs)
 - `saveFileAtom` actually writes the content of the buffer on the disk. It does NOT update the `loadableFileAtom`, to prevent the editor from re-rendering and resetting focus and history.

Closes #197


src/atoms/fileActions.test.ts
```diff
@@ -463,4 +471,148 @@ describe('fileActions', () => {
...
+    const paneToFocus: PaneId = activePane.current.id === 'LEFT' ? 'RIGHT' : 'LEFT';
+
+    act(() => {
+      focusPanel.current(paneToFocus);
+    });
+
+    expect(activePane.current.id).toBe(paneToFocus);
+  });
+
+  describe('File content atoms', () => {
```
cguedes, line 489: nit: maybe we should split this file in multiple files by scope.
For example, you can see that we have two afterEach in this file :-).

cguedes (approved):

src/atoms/fileActions.test.ts
sehyod: I agree but I don't know how to name the files, because so far we have been using the convention `fileName.test.ts` to test `fileName.ts`. The issue is that `fileActions.ts` contains a lot of logic. Do you think we should split it too?

For the second `afterEach`, it's because it's in a nested `describe` block, but apparently top-level `afterEach` also runs after each nested test, so you're right, I can remove it

sehyod: I agree but I don't know how to name the files, because so far we have been using the convention fileName.test.ts to test fileName.ts. The issue is that fileActions.ts contains a lot of logic. Do you think we should split it too?

For the second afterEach, it's because it's in a nested describe block, but apparently top-level afterEach also runs after each nested test, so you're right, I can remove it

cguedes: You are right, maybe we should also split `fileActions.ts`. We can make that change in a different PR if you want.

cguedes (dismissed):

cguedes (approved):

danvk:
Exciting that we can persist files now! Two questions:

- Is there a "factory reset" option available? I was playing around with Tiptap and got myself into a bad state. Running `yarn tauri:dev:debug` no longer resets my state, so I'm not sure how to get out of this.
- When you save a file to disk, where does it go? The `refstudio` directory?

sehyod:
- There is no such option for now. However, we are planning on adding the ability to remove files (see #212). Could you give more details about the bad state you got yourself into? Maybe it's worth filing an issue about it?
- It goes to the project subdirectory in the `refstudio` directory. Something like `/Users/$USER/Library/Application Support/com.tauri.dev/project-x`

cguedes:
@danvk we've had to update the references model (FE), and since we persist that information in the localStorage sometimes you need to reset that info using the UI (i.e. "DEBUG: reset references store") or directly via dev tools (Application > Storage). I normally also run the `./scripts/reset_uploads.sh` utility script to reset all the BE data.


image

PR #217, opened 2023-06-28
feat: Add sidecar endpoint so UI can pass reference metadata updates to backend

### LINKED ISSUE #182: Add sidecar endpoint so UI can pass Reference metadata updates to backend

gjreda:
fixes #182 
______

```bash
# Full example showing the update
$ cat .storage/references.json | jq '.[] | select(.source_filename | contains("fails")) | {"source_filename": .source_filename, "title": .title}'

{
  "source_filename": "grobid-fails.pdf",
  "title": null
}

# run the update
$ poetry run python main.py update --data '{"source_filename": "grobid-fails.pdf", "patch": {"title": "a title that was missing"}}'

{
  "status": "ok",
  "message": ""
}

# check that it worked
$ cat .storage/references.json | jq '.[] | select(.source_filename | contains("fails")) | {"source_filename": .source_filename, "title": .title}'

{
  "source_filename": "grobid-fails.pdf",
  "title": "a title that was missing"
}
```


sehyod:
For 1, I think it makes sense to interpret empty strings as null. I don't have any strong opinion on whether it is the frontend's job to send null instead of an empty string or if the backend should convert empty strings to null as you suggested.

And for 2, I think it's actually easier for the frontend too to send the whole array of authors to the backend rather than keeping track of what indexes have been updated

cguedes:
1. I'm ok with sending empty strings, bc most times that's what the UI can do/detect for when a field is cleared (one just remove all the text). 
2. That's ok.

cguedes (approved):

PR #227, opened 2023-06-29
Remove filename_md5 from backend

### LINKED ISSUE #190: chore: Get rid of filename_md5 field on Reference
In chatting with @cguedes, we decided this field isn't really necessary. For now, we can consider the `source_filename` to be the unique identifier of the paper, since the filesystem will require it to be unique anyway.

Longer term we should probably so something like a UUID or database primary key. @shauryr mentioned that S2 uses a SHA1 of the PDF as an identified, so we could also adopt similar.

gjreda:
fixes #190 
____
Discussed this with @cguedes and it makes more sense for us to consider the `source_filename` to be the true unique id of the Reference file, since that implicitly happens on the filesystem anyway.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -63,7 +63,6 @@ $ poetry run python main.py ingest --pdf_directory=tests/fixtures/pdf/ | jq
   "references": [
     {
       "source_filename": "test.pdf",
-      "filename_md5": "754dc77d28e62763c4916970d595a10f",
       "title": "A Few Useful Things to Know about Machine Learning",
       "abstract": "... PDF abstract here ...",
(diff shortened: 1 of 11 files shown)


gjreda:
@cguedes or @sehyod can y'all help with the front-end changes for this? It looks like we use `filename_md5` in a handful of places and I felt unsure about refactoring that.

sehyod:
I have pushed the changes for the frontend:
- updated `parsePdfIngestionResponse` to use `source_filename` as id
- removed `filename_md5` from test fixtures

cguedes (approved):

PR #241, opened 2023-07-02
feat: Add source reference metadata to chat response

### LINKED ISSUE #143: Add source Reference metadata to Chat response
A Chat response currently only includes the message text. Ideally, we'd like to cite the References that were used to create the message response.

We should update the response to include metadata about the source References.

- [x] Update ingest to use `pypdf` for text extraction on _all_ PDFs so that text for grobid success/fails are parsed in a consistent way. This will allow us to also include page numbers on our Reference metadata.
- [x] Add page number to `Chunk` metadata
- [x] Add reference metadata [to ranker context here](https://github.com/refstudio/refstudio/blob/main/python/sidecar/ranker.py#L18) so that our ranker query returns appropriate metadata
- [x] Update chat to include relevant reference metadata in in prompt
- [x] Update LLM response to include citations of appropriate references (might need to play around with LLM prompt)
- [x] Update CLI params to allow for `n_choices` to be utilized
- [ ] Update python/readme.md

gjreda:
fixes #143 
_____

Some examples below.

Note that in the first example, the LLM appropriately cities multiple sources. I was surprised that it nicely combined them without an example or being told how to do so.
```bash
$ poetry run python main.py chat --text "What are some types of technical debt in machine learning?" | jq
[
  {
    "index": 0,
    "text": "Some types of technical debt in machine learning include erosion, entanglement, hidden feedback loops, undeclared consumers, data dependencies, changes in the external world, system-level anti-patterns, configuration issues, and code complexity issues.",
    "citations": [
      {
        "source_filename": "Machine Learning - The High-Interest Credit Card of Technical Debt.pdf",
        "page_num": 1
      },
      {
        "source_filename": "Hidden Technical Debt in Machine Learning Systems.pdf",
        "page_num": 1
      }
    ]
  }
]

$ poetry run python main.py chat --text "What are hidden feedback loops?" | jq
[
  {
    "index": 0,
    "text": "Hidden feedback loops are systems that influence each other indirectly through the world, often resulting in changes in behavior in reaction to changes in one system. These loops may exist between completely disjoint systems.",
    "citations": [
      {
        "source_filename": "Hidden Technical Debt in Machine Learning Systems.pdf",
        "page_num": 4
      },
      {
        "source_filename": "Machine Learning - The High-Interest Credit Card of Technical Debt.pdf",
        "page_num": 2
      }
    ]
  }
]

$ poetry run python main.py chat --text "What can you tell me about Chicago?" | jq
[
  {
    "index": 0,
    "text": "I am unable to answer the question with the provided sources.",
    "citations": []
  }
]


sehyod (dismissed):
Nice! This greatly improves the chat experience!

cguedes (approved):
This is a great step forward working with the chat, but my expectation was that we would get structured data reply with the reference(s) in a field outside of the text. Can we have it @gjreda?

gjreda:
> This is a great step forward working with the chat, but my expectation was that we would get structured data reply with the reference(s) in a field outside of the text. Can we have it @gjreda?

@cguedes updated this to include structured citations. see the updated top comment on this PR for examples

sehyod:
I'm sorry about the back-and-forth regarding this feature, but I'm a bit concerned about this update, which feels like regression to me.

The reply we were getting before for the question about hidden feedback loops was the following:
`Hidden feedback loops in machine learning refer to situations where two systems indirectly influence each other through the world. One example of this is when two systems independently determine different aspects of a web page, such as selecting products to show and selecting related reviews. Improving one system may lead to changes in behavior in the other, as users react to the changes by clicking more or less on the other components (Hidden Technical Debt in Machine Learning Systems.pdf, p4). These hidden feedback loops can exist between completely disjoint systems and pose a statistical challenge for ML researchers to investigate (Machine Learning - The High-Interest Credit Card of Technical Debt.pdf, p3).`

The new reply we get is:
```Hidden feedback loops are systems that influence each other indirectly through the world, often resulting in changes in behavior in reaction to changes in one system. These loops may exist between completely disjoint systems.```

Although we now get citations with a proper schema, I think the reply contains far less details. And another issue is that we don't know which part of the answer corresponds to which citation.
Maybe we should rollback to the previous solution and parse content between parenthesis with a regex to retrieve which references are being used?
What are your thoughts on that?

cguedes:
I'm ok with reverting to the other solution and regex/parse the reference for now and work on the improved solution in a different PR.

gjreda:
Yes, requiring JSON output means less available token space in the response, and it also means eating up a lot more of the input context space due to requiring more directions in the prompt. It took a lot of trial and error to have a prompt that resulted in JSON output and also left enough of the input context space for the relevant text chunks.

I don't love the idea of reworking this again and trying to regex out a somewhat non-deterministic result from the LLM, but can give it a shot if we think it's worth the time.

I think the best approach for now would be to move forward with how this PR previously was (with inline citations) and then I can add a follow-up PR to regex out the citations if we think it's valuable.

cguedes:
@gjreda we see your point. Ok. Push a revert to this branch and let me know when it's ready for a new review.

gjreda:
@cguedes Sound good, should be good to go now and I'll work on a separate regex PR.

cguedes (approved):

PR #243, opened 2023-07-04
feat: add ability to delete file

### LINKED ISSUE #212: Add ability to delete files

sehyod:
This adds a context menu to files in the file explorer.
The menu has one action: `Delete`, that removes a file

Closes #212


src/application/App.tsx
```diff
@@ -16,6 +18,7 @@ import { ApplicationFrame } from '../wrappers/ApplicationFrame';
 import { EventsListener } from '../wrappers/EventsListener';
 import { MainPanel } from './components/MainPanel';
 import { ExplorerPanel } from './sidebar/ExplorerPanel';
+import { FileExplorerContextMenus } from './sidebar/fileExplorerContextMenu/FileExplorerContextMenus';
```
cguedes, line 21: Don't you think we should move this to wrappers instead?

src/application/sidebar/fileExplorerContextMenu/fileExplorerContextMenu.css
```diff
@@ -0,0 +1,15 @@
+body {
+  --contexify-menu-bgColor: #f1f5f9;
+  --contexify-item-color: #000000;
+  --contexify-activeItem-color: #000000;
+  --contexify-activeItem-bgColor: #e2e8f0;
+}
+
+.contextMenu {
+  @apply p-1;
+  min-width: min-content;
```
cguedes, line 10: you can use `@apply min-w-min`

src/io/filesystem.ts
```diff
@@ -150,3 +151,14 @@ function convertTauriFileEntryToFileEntry(entry: TauriFileEntry, baseDir: string
     };
   }
 }
+
+export async function deleteFile(relativePath: string): Promise<boolean> {
+  const path = (await getBaseDir()) + relativePath;
```
cguedes, line 156: use `join` instead

src/settings/SettingsModal.tsx
```diff
@@ -61,7 +61,7 @@ export function SettingsModal({ open, onCloseClick }: { open: boolean; onCloseCl
   }
 
   return (
-    <div className="fixed left-0 top-0 flex h-screen w-screen items-center justify-center">
+    <div className="z-99 fixed left-0 top-0 flex h-screen w-screen items-center justify-center">
```
cguedes, line 64: my PR has `z-[99999]` :)

cguedes (changes requested):

src/io/filesystem.ts
sehyod: Thank you!

src/settings/SettingsModal.tsx
sehyod: I didn't mean to add this to this PR, I had changed the code to fix the interface while I was testing it, but then accidentally added it to my commit. I will revert it, to avoid conflicts

PR #245, opened 2023-07-05
fix: remove footer border

sehyod:
This removes the top border of the footer and the bottom border of the collapsed AI panel

Before:
image

After:
image


cguedes:
I think that if we remove the border we need to make the left sidebar non-black.

image

image

image

cguedes (approved):

PR #247, opened 2023-07-05
fix: open pdf from references with filename

### LINKED ISSUE #246: Opening PDF files from the references panel does no longer work
#244 introduced a regression bug: the app expects relative paths everywhere, but the path returned by the backend for references is the absolute path. When trying to open the PDF corresponding to this ref, the app tries to access an invalid path, where the base directory is duplicated: `/Users/mathis/Library/Application Support/com.tauri.dev/project-x/Users/mathis/Library/Application Support/com.tauri.dev/project-x/uploads/Software Engineering Practices for Machine Learning.pdf` instead of `/Users/mathis/Library/Application Support/com.tauri.dev/project-x/uploads/Software Engineering Practices for Machine Learning.pdf` for example.

sehyod:
Fixes #246 

This creates a new atom to open the PDF file corresponding to a reference. The atom checks that the reference exists and then call the `openFilePathAtom`, assuming that the PDF lives in the uploads directory

### DIFF
--- a/src/features/references/__tests__/test-fixtures.ts
+++ b/src/features/references/__tests__/test-fixtures.ts
@@ -1,9 +1,10 @@
+import { UPLOADS_DIR } from '../../../io/filesystem';
 import { ReferenceItem } from '../../../types/ReferenceItem';
 
 export const REFERENCES: ReferenceItem[] = [
   {
     id: 'a_few_useful_things_to_know_about_machine_learning',
-    filepath: '/path/to/A Few Useful Things to Know about Machine Learning.pdf',
+    filepath: `/${UPLOADS_DIR}/A Few Useful Things to Know about Machine Learning.pdf`,
     filename: 'A Few Useful Things to Know about Machine Learning.pdf',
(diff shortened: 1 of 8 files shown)


src/features/references/__tests__/test-fixtures.ts
```diff
@@ -1,9 +1,10 @@
+import { UPLOADS_DIR } from '../../../io/filesystem';
 import { ReferenceItem } from '../../../types/ReferenceItem';
 
 export const REFERENCES: ReferenceItem[] = [
   {
     id: 'a_few_useful_things_to_know_about_machine_learning',
-    filepath: '/path/to/A Few Useful Things to Know about Machine Learning.pdf',
+    filepath: `/${UPLOADS_DIR}/A Few Useful Things to Know about Machine Learning.pdf`,
```
cguedes, line 7: We shouldn't need this. Do we really need this dependency on the filesystem here?

cguedes (approved):

PR #250, opened 2023-07-05
138 basic search functionality

shauryr:
This pull request contains a new script `search.py` that allows users to query Semantic Scholar. The search can be customized via command line arguments to specify a search query and the number of results desired.

Key features:
- The script creates a wrapper class `Search` around the Semantic Scholar search process.
- I've added a method to preprocess the search query, which includes converting to lowercase and removing stopwords.
- The search result is a list of dictionaries where each dictionary contains key information about a paper.

The script outputs a list of dictionaries where each dictionary represents a paper and contains key-value pairs for paper variables such as title, abstract, and venue, among others. 

Specifically, JSON object structure for each paper is as follows:

```json
{
    "title": "<TITLE>",
    "abstract": "<ABSTRACT>",
    "venue": "<VENUE>",
    "year": "<YEAR>",
    "corpusId": "<CORPUSID>", # this is useful if we want to fetch other information about the paper later
    "citationCount": "<CITATIONCOUNT>",
    "openAccessPdf": "<LINK TO OPEN ACCESS PDF HERE>"
}
```

I've thoroughly tested the script and confirmed that the output matches expectations. The `main` module of the script demonstrates its usage.

Please review these changes and let me know about any feedback, questions, or concerns you might have.


python/sidecar/search.py
```diff
@@ -0,0 +1,256 @@
...
+import argparse
+from semanticscholar import SemanticScholar
+
+
+class Search:
+    # Initialize the SemanticScholar API client and stopwords set
+    def __init__(self):
+        self.s2 = SemanticScholar()
+        self.stopwords = set(
+            [
```
cguedes, line 22: @shauryr can you move these stopwords to a different file?

python/sidecar/search.py
```diff
@@ -0,0 +1,256 @@
...
+        # For each result, create a dictionary with desired fields
+        for item in results[:limit]:
+            result = {field: getattr(item, field, None) for field in search_fields}
+            results_list.append(result)
+        return results_list
+
+
+# Main function to parse arguments and perform the search
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Semantic Scholar search")
```
cguedes, line 242: The parser is configured in the ``cli.py` file and not here.

cguedes (changes requested):
I'm not sure if it's out-of-scope for this PR, but I don't see the integration of this functionality as a new operation for the backend as I expected. It's still missing:

- main.py code
- output types
- unit tests
- README

gjreda:
@shauryr Nice! We'll need to make some changes before merging, but this is a great start. I'm dropping some notes below, but I'm also happy to chat about this in detail tomorrow.

I also think it'd be good to iron out the "what's next" for search on tomorrow's weekly call -- what can the user do with the search results that are displayed to them?

___

The python sidecar is effectively one big command line application with various different subcommands like `ingest`, `chat`, `rewrite`, etc.

To allow the Tauri front-end to use this new search functionality, we'll want to add a new subcommand for it to the sidecar app. 

We'll need to:
- [ ] Add a `search` subparser to `cli.py` (see [rewrite as an example here](https://github.com/refstudio/refstudio/blob/main/python/sidecar/cli.py#L37)).
- [ ] Update `main.py` with the new command (see [rewrite as an example here](https://github.com/refstudio/refstudio/blob/main/python/main.py#L21))
- [ ] Add some tests (we can mock the call to the S2 API, much [like we do with OpenAI now](https://github.com/refstudio/refstudio/blob/main/python/tests/test_rewrite.py#L7:L31))

Additionally, the sidecar and front-end communicate with each other by passing `json` back and forth via stdout, so we'll need to both read from stdin for this and write search results back to stdout. [Here's an example where we read from stdin](https://github.com/refstudio/refstudio/blob/main/python/main.py#L27) to [update the metadata of a reference](https://github.com/refstudio/refstudio/blob/main/python/sidecar/storage.py#L10). And [here's an example where we pass data back to the UI](https://github.com/refstudio/refstudio/blob/main/python/sidecar/ingest.py#L33).

python/sidecar/search.py
```diff
@@ -0,0 +1,256 @@
...
+Instructions:
+The script takes two command line arguments:
+--query: The term you want to search on Semantic Scholar, written in quotation marks (e.g., "brain computer interface").
+--limit: Optional, the maximum number of papers to retrieve. Default is 10.
+
+Run the script like so from your terminal:
+    python search.py --query "computational linguistics" --limit 10
+"""
+import argparse
+from semanticscholar import SemanticScholar
```
gjreda, line 14: We'll need to add this as a dependency via `poetry add semanticscholar` and adding the resulting update to pyproject.toml and poetry.lock.

python/sidecar/search.py
gjreda: Agree - we can just have a `stopwords.py` in the sidecar directory

python/sidecar/search.py
```diff
@@ -0,0 +1,256 @@
...
+        help="Search query for the Semantic Scholar search",
+    )
+    parser.add_argument(
+        "--limit", type=int, default=10, help="Limit for the number of search results"
+    )
+    args = parser.parse_args()
+
+    ss = Search()
+    results = ss.search(args.query, args.limit)
+    print(results)
```
gjreda, line 256: We'll need to emit these as json via `sys.stdout.write(results)`

python/sidecar/search.py
```diff
@@ -0,0 +1,256 @@
...
+            "openAccessPdf",
+        ],
+    ):
+        # Preprocess the query
+        query = self.preprocess_query(query)
+        results = self.s2.search_paper(query, limit=limit, fields=search_fields)
+        results_list = []
+        # For each result, create a dictionary with desired fields
+        for item in results[:limit]:
+            result = {field: getattr(item, field, None) for field in search_fields}
```
gjreda, line 235: We should make the results an explicit type (see examples [here](https://github.com/refstudio/refstudio/blob/main/python/sidecar/typing.py)). Maybe a `S2SearchResult` type? What do you think @shauryr @cguedes

gjreda (changes requested):

python/sidecar/search.py
```diff
@@ -0,0 +1,256 @@
...
+        query,
+        limit=10,
+        search_fields=[
+            "title",
+            "abstract",
+            "venue",
+            "year",
+            "corpusId",
+            "citationCount",
+            "openAccessPdf",
```
cguedes, line 226: @shauryr are these output fields or search fields? Also, can we have authors?

python/sidecar/search.py
cguedes: I think that the response type should be agnostic of being a scholar paper search. In addition we should maybe have a field for the search provider (now Scholar) and trying to have the output field names not using s2 specific fields (ex: openAccessPdf -> pdfUrl`

I've noticed that the `paperId` is the primary id, and that `corpusId` is a second way to identify papers. Should we keep both or one of them (`paperId` imo)

ex of using the paperId to get paper details: https://api.semanticscholar.org/api-docs/#tag/Paper-Data/operation/get_graph_get_paper

cguedes (changes requested):
Added some additional notes for this feature.

python/sidecar/search.py
shauryr: Great suggestion! we can just keep one id, `paperId` @cguedes 
`paperId` will also be the same as the SHA1 of a pdf file (sometimes)

python/sidecar/search.py
shauryr: these are output fields. Yes, I will add authors as well.

shauryr: authors added - https://github.com/refstudio/refstudio/pull/250/commits/26b9f12a134ffafd2b81a31a1ba9246c5d140331

python/sidecar/search.py
shauryr: `stopwords.py` has been added - https://github.com/refstudio/refstudio/pull/250/commits/d089fa182f698b0bd254382c5eece50255ba5074

python/main.py
```diff
@@ -24,6 +24,9 @@
     if args.command == "chat":
         chat.ask_question(args.text)
     
+    if args.command == "search":
+        search.search(args.query)
```
gjreda, line 28:
```suggestion
        search.search(query=args.query, limit=args.limit)
```

python/sidecar/search.py
```diff
@@ -0,0 +1,65 @@
...
+    for item in results[:limit]:
+        # For each result, create a dictionary with the fields specified in returned_fields
+        result = {field: getattr(item, field, None) for field in returned_fields}
+        # Replace author dictionaries with author names in each result
+        if 'authors' in result:
+            result['authors'] = [author['name'] for author in result['authors']]
+        results_list.append(result)  # add the result to the list of results
+    
+    # Convert the results_list to JSON and write to stdout
+    sys.stdout.write(json.dumps(results_list))
```
gjreda, line 65: We should use an explicit type here. Probably named something like `S2SearchResult`

gjreda: We should pass a list of `S2SearchResult`.

```python
class S2SearchResult(RefStudioModel):
    # fields the same as you already have in dict
    # just explicitly declared
```

And search should emit a type of `SearchResponse`

```python
class SearchResponse(RefStudioModel):
    status: ResponseStatus
    results: list[S2SearchResult]
    
response = SearchResponse(status=ResponseStatus.OK, results=results_list)
```
And then at `sys.stdout.write` we can do `sys.stdout.write(response.json())`

python/tests/test_search.py
```diff
@@ -0,0 +1,46 @@
+from sidecar import search
+
+def test_search(monkeypatch):
```
gjreda, line 3:
```suggestion
def test_search(monkeypatch, capsys):
```

python/tests/test_search.py
```diff
@@ -0,0 +1,46 @@
...
+                    {"authorId": "auth-id-4", "name": "Sample Author Name 4"},
+                ]
+            },
+        ]
+        return response
+
+    monkeypatch.setattr(search.Search, "search_paper", mock_search_paper)
+
+    searcher = search.Searcher()
+    result = searcher.s2.search_paper("any-query-string-you-like")
```
gjreda, line 38: Change this to test `search.search`

python/sidecar/search.py
```diff
@@ -0,0 +1,65 @@
...
+    This function performs a search on Semantic Scholar's API and 
+    writes the results in JSON to stdout.
+    '''
+
+    # Create a Search object
+    search = Searcher()
+    # Preprocess the query (e.g., lowercase, remove stopwords)
+    query = search.preprocess_query(query)
+    # Perform the search using Semantic Scholar's API
+    results = search.s2.search_paper(query, limit=limit, fields=returned_fields)
```
gjreda, line 52: Maybe add some `log` statements here

gjreda: See examples in `ingest.py`

python/sidecar/search.py
shauryr: changes made here - 8e3fdf39e92caad43c3091887acfd2d518e00b25

python/tests/test_search.py
shauryr: resolved in https://github.com/refstudio/refstudio/pull/250/commits/b716691359753cba88ef5d0ef28f3b73e494fe8c

python/tests/test_search.py
shauryr: resolved in https://github.com/refstudio/refstudio/pull/250/commits/b716691359753cba88ef5d0ef28f3b73e494fe8c

python/sidecar/search.py
shauryr: added logging - https://github.com/refstudio/refstudio/pull/250/commits/faac256ca624e3846a6d907e64e259863d7afffb

python/cli.schema.json
```diff
@@ -582,6 +583,21 @@
...
       ]
+    },
+    "SearchResponse": {
+      "title": "SearchResponse",
+      "type": "object",
+      "properties": {
+        "search_results": {
+          "type": "array",
+          "items": {
+            "type": "object"
```
cguedes, line 594: @shauryr can you make this a typed reference?
Also, you should run the `yarn codegen` to regenerate the `src/api/types.ts` definitions for typescript.

python/sidecar/cli.py
```diff
@@ -10,16 +10,75 @@ def get_arg_parser():
         "--debug",
         action="store_true",
     )
-    parser.add_argument(
-        'command',
-        # To add a new Sidecar command, add it to the CliCommands class.
-        choices=[*inspect.signature(CliCommands).parameters.keys()],
+    subparsers = parser.add_subparsers(
```
cguedes, line 13: @shauryr why have you changed this parser call to use `add_subparsers`?

PR #260, opened 2023-07-06
TypeScript types for sidecar CLI args

danvk:
Follow-on to #126 to generate TypeScript types for the CLI args for the sidecar (this temporary solution has had more staying power than we expected!).

The flags are only specified via interactions with `ArgumentParser`. This has to go through some internal APIs and make some assumptions about what's required/optional to work.

An alternative approach would be to make all the subcommands take a single JSON argument and then we could use Pydantic to model that, same as we do with outputs. This would be a lot simpler/cleaner, but it would make sidecar invocations from the command line less idiomatic, i.e.

    python cli.py ingest '{"pdf_directory": "path/to/pdfs"}'

vs the current:

    python cli.py ingest --pdf_directory path/to/pdfs

Would love some input from @gjreda on whether this would be annoying or if we should just embrace JSON.


src/api/__tests__/chat.test.ts
```diff
@@ -1,15 +1,15 @@
 import { chatWithAI } from '../chat';
-import { callSidecar } from '../sidecar';
+import { callSidecarWithArgs } from '../sidecar-with-args';
 
-vi.mock('../sidecar');
```
danvk, line 4: `vi.mock('module')` mocks all the functions in the module. We only want it to mock one (`callSidecarWithArgs`, not `callSidecar`). The easiest way is to mock just one function in a module is to just move it into a standalone module. https://stackoverflow.com/q/59312671/388951

python/generate_schema.py
```diff
@@ -17,4 +27,39 @@
...
+        assert help.dest == 'help'
+        ts += f'/* {description} */\n'
+        ts += f'export interface {ts_name} {{\n'
+        for arg in args:
+            if arg.help:
+                ts += f'  /* {arg.help} */\n'
+            option_name = arg.option_strings[0].replace('--', '')
+            mod = ''
+            typ_ = 'string' if arg.type == str else 'number' if arg.type in {int, float} else 'unknown'
+            if isinstance(arg, argparse._StoreTrueAction):
```
danvk, line 48: I don't love all this code. See my idea in the PR description about making the CLI take JSON instead.

src/api/__tests__/chat.test.ts
danvk: (this is irrelevant after the refactor to use Pydantic)

python/generate_schema.py
danvk: This gross code is gone!

python/README.md
```diff
@@ -273,17 +280,18 @@ $ poetry run python main.py delete --source_filenames file-does-not-exist.pdf |
...
 `Update` updates a Reference's metadata in project storage. It is called when a user updates details about a Reference.
 
-It takes an input argument of `--data` which is a dictionary containing a key for `source_filename` and `patch`. `source_filename` acts as the unique ID for the Reference to update and `patch` is a dictionary containing the corresponding fields and values to be used for the update. 
+Its input argument has a key for `source_filename` and `patch`. `source_filename` acts as the unique ID for the Reference to update and `patch` is a dictionary containing the corresponding fields and values to be used for the update.
 
 To run `update`:
 
 ```bash
-$ poetry run python main.py update --data '{"source_filename": "grobid-fails.pdf", "patch": {"title": "a title that was missing"}}'
+$ poetry run python main.py update '{"source_filename": "grobid-fails.pdf", "patch": {"title": "a title that was missing"}}'
```
danvk, line 291: Looks like `update` had already hit on the one-arg JSON form :)

gjreda (approved):
LGTM, but let's hold off on merging until next week since merging this will cause some conflicts to other PRs. Once the other PRs get merged, I will resolve the merge conflicts on this one.

cguedes (approved):

cguedes:
> LGTM, but let's hold off on merging until next week since merging this will cause some conflicts to other PRs. Once the other PRs get merged, I will resolve the merge conflicts on this one.

Ok. Will make this a draft PR.

gjreda:
Resolved the expected merge conflicts. For the sidecar, I'm ok with us merging this now. What do you think @cguedes?

gjreda (approved):

cguedes:
@gjreda you can merge it 💪

PR #267, opened 2023-07-07
feat: notification system and UI

### LINKED ISSUE #263: Notification system and UI
The current version of RefStudio lacks a robust notification system and user interface (UI) for displaying notifications to users. This issue aims to enhance the application by adding support for notifications and improving the associated UI.

The absence of a notification system hampers the ability of our application to deliver timely updates and important information to users. Users may miss out on critical updates, such as new messages, reminders, or system alerts, resulting in a suboptimal user experience.

cguedes:
This PR closes #263 

Add notification system for RefStudio that can display info, warning and error notifications.

The notifications are triggered (and cleared) using a simple API:

```js
notifyInfo('Message', 'Optional message details`);
notifyWarning('Message', 'Optional message details`);
notifyError('Message', 'Optional message details`);
notifyErr(someErr, 'optional title override');

clearNotifications(type?);
```

The API is an utility wrapper that trigger the event `refstudio://notifications/new` and `refstudio://notifications/clear` with a payload.

The `EventsListener` component registers two _hook listeners_ that transform events into calls to the notifications store.

The **`notificationsState.ts`** holds the state for notifications. This state is a sorted list of notification items containing:

```js
export type NotificationItemType = 'info' | 'warning' | 'error';

export interface MutableNotificationItem {
  type: NotificationItemType;
  id: string;
  when: string;
  title: string;
  details?: string;
}

export type NotificationItem = Readonly<MutableNotificationItem>;
```
The `MutableNotificationItem` is used internally in the Jotai store and the readonly `NotificationItem` is the public type.

For the UI, this PR adds up to 4 footer items in the right side.

| without notifications | with all notifications (including error) |
|--------|--------|
| image | image |

The three first footer items show the number of notifications of type: `info`, `warning` and `error`. They will be hidden if no notification (of that type) exists. The fourth (🔔) is always visible and displays a dot if a notification exists. Additionally, it displays with a red background if there is one or more notifications of type error.

Clicks on the footer items show a notification items popup with details about the notifications. If the latest (top) notification has details, the details are expanded by details.

Additionally, if a notification of type error occurs, the notifications popup opens automatically.

A system menu (`View > Notifications (F11)`) was also implemented in order to show/hide the notifications popup.

image

----

During development is common to write in the console useful information (info, warning, even errors) that one don't necessarily want to be deployed or visible in the final deployment.

This PR adds an utility function (`interceptConsoleMessages(bool, bool, bool)`) that intercepts `console.{info,warn,error}` calls adding a call to the corresponding `notity{Info,Warning,Error,Err}` method. The functionality can be turned on/off for each type of notification and also globally disabled. In the future we might want to connect this feature to settings.

----

extra:
  - reorganize the placement of references loading (footer -> app)
  - Adds new utility test script `yarn test:watch:ui`


src/notifications/__tests__/notifications.console.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    expect(vi.mocked(notifyInfo)).toHaveBeenCalledTimes(1);
+    expect(vi.mocked(notifyInfo)).toHaveBeenCalledWith(prefixTitle('message'), '"some extra"');
+  });
+
+  it('should intercept console.log with optional parameters in multi-line JSON', () => {
+    console.log('message', 'some extra', 1, { a: 42 });
+    expect(vi.mocked(notifyInfo)).toHaveBeenCalledTimes(1);
+    expect(vi.mocked(notifyInfo)).toHaveBeenCalledWith(
+      prefixTitle('message'),
+      ['"some extra"', 1, JSON.stringify({ a: 42 }, null, 2)].join('\n'),
```
sehyod, line 37: Where are `null` and `2` coming from?

src/notifications/notifications.console.ts
```diff
@@ -0,0 +1,66 @@
...
+  console.error = interceptError
+    ? (message?: unknown, ...optionalParams: unknown[]) => {
+        const [firstOptional] = optionalParams;
+
+        if (message instanceof Error) {
+          // Sometimes the `message` is an Error.
+          notifyErr(message);
+        } else if (firstOptional instanceof Error) {
+          // Sometimes the `firstOptional` is an Error. Let's use that and override the title
+          notifyErr(firstOptional, prefixTitle(message));
```
sehyod, line 41: What does `message` correspond to when `firstOptional` is an error?

src/notifications/notifications.console.ts
```diff
@@ -0,0 +1,66 @@
...
+    console.error = error;
+  };
+}
+
+function formatOptionalParams(optionalParams: unknown[]) {
+  if (optionalParams.length === 0) {
+    return undefined;
+  }
+
+  const output = optionalParams.map((param) => JSON.stringify(param, null, 2)).join('\n');
```
sehyod, line 64: Why `null, 2`?

src/notifications/notifications.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function notifyWarning(title: string, details?: string) {
+  emitEvent('refstudio://notifications/new', { type: 'warning', title, details });
+}
+
+export function notifyError(title: string, details?: string) {
+  emitEvent('refstudio://notifications/new', { type: 'error', title, details });
+}
+
+export function notifyErr(err: unknown, title?: string, details?: string) {
```
sehyod, line 16: I think the type should be `err: Error` here

sehyod:
This is very nice!
A few extra comments after having tested the feature locally:
 - F11 does not work for me, I think it's already a MacOS shortcut, to reveal the desktop
 - z-index for the notifications is not high enough, so it appears below the AI panel

src/notifications/notifications.console.ts
cguedes: To format with 2 spaces (and no replacer). We will use that text in the UI.

src/notifications/footer/FooterNotificationItems.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const notifications = useAtomValue(listNotificationsAtom);
+  const stats = useAtomValue(notificationsTypeStatsAtom);
+
+  const notificationsPopup = useAtomValue(notificationsPopupAtom);
+
+  useEffect(() => {
+    if (notifications.length > 0 && notifications[0].type === 'error') {
+      showNotifications('error');
+    }
+  }, [notifications]);
```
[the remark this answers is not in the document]
cguedes: This is the UX in VS Code and I think it would be usefull. Will keep it for now and then we can hide if is anying.

src/notifications/notifications.console.ts
cguedes: Will be the title. Sometimes we use this pattern: `console.error('Error calling chat', err);`

src/notifications/footer/FooterNotificationItems.tsx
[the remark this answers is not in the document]
sehyod: My main concern is that we have errors when switching between editors, because of TipTap (which by the way has been fixed in a recent PR: https://github.com/ueberdosis/tiptap/pull/4000), so the the drawer will open a lot

src/notifications/footer/NotificationsPopup.tsx
```diff
@@ -35,7 +35,7 @@ export function NotificationsPopup({
     </div>
   );
 }
-function NotificationItemWidget({ item, expanded }: { item: Readonly<NotificationItem>; expanded?: boolean }) {
+function NotificationItemWidget({ item, expanded }: { item: Readonly<ReadonlyNotificationItem>; expanded?: boolean }) {
```
sehyod, line 38: `Readonly` is duplicated here

sehyod (approved):

PR #273, opened 2023-07-10
Rewrite filesystem and add unit tests

cguedes:


src/io/filesystem.sample-content.ts
```diff
@@ -0,0 +1,15 @@
+export { INITIAL_CONTENT } from '../features/textEditor/components/tipTapEditorConfigs';
```
sehyod, line 1: I think you move `INITIAL_CONTENT` here, there is no need to keep it in `tipTapEditorConfigs`.

sehyod (dismissed):

sehyod (approved):

Issue #261, opened 2023-07-06
Update tauri bundle identifier


danvk:
We're still using the Tauri defaults:

https://github.com/refstudio/refstudio/blob/ed6880d3e6de54f5908b1800bf0e57eea0e4eb7e/src-tauri/tauri.conf.json#L45

ideally this would be something like `com.refstudio` or `studio.ref` or anything that we control. Since we store all files in a directory that includes this identifier, changing this would be like doing a data reset for everyone.

cguedes:
Agree. What should we use @hammer?

hammer:
@cguedes we own ref.studio. Should we use that or is it advisable to put a subdomain onto it?

cguedes:
I think we should have a subdomain (app.ref.studio) there like: `studio.ref.app` so that we don't "use" the `ref.studio` for this.

Related [discussion](https://github.com/tauri-apps/tauri/discussions/5976) about this topic.

PR #296, opened 2023-07-13
feat: update tauri bundle identifier

cguedes:
This will close https://github.com/refstudio/refstudio/issues/261

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -9,6 +9,9 @@ fn get_environment_variable(name: &str) -> String {
 
 use dotenv::dotenv;
 use std::env;
+
+// Prevent build warning for tauri::Manager
+#[allow(unused_imports)]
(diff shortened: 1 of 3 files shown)


src-tauri/src/main.rs
```diff
@@ -19,12 +22,12 @@ fn main() {
     let context = tauri::generate_context!();
 
     tauri::Builder::default()
-        .setup(|app| {
+        .setup(|_app| {
```
sehyod, line 25: Why do we need to rename this?

cguedes: That's a weird warning during `yarn tauri build` for unused variable 😕.
I didn't spend time understanding what is going on (is not that important for now, IMO), so just followed the compiler instructions to disable the warning.

sehyod (approved):

PR #302, opened 2023-07-13
feat: Handle errors, retries, and improve response types

### LINKED ISSUE #85: Add error and rate limit handling for calls to OpenAI API
We'll have a better sense of how we want to handle this once we have some UX for the "summarize" AI interactions.

- https://github.com/openai/openai-cookbook/blob/main/examples/How_to_handle_rate_limits.ipynb

gjreda:
@cguedes @sehyod I pushed this with `no-verify` since I figured we might have some discussion on the typing changes. Leaving as draft for now, but please feel free to add your feedback.
______
fixes #85 and allows errors to be return within the sidecar response types

For now, I'm passing through the exception in the `message`, but I figured this PR would lead to some follow-up discussion on what we'd want to see on errors.
_____

*Ok*
```
{
  "status": "ok",
  "message": "",
  "choices": [
    {
      "index": 0,
      "text": "ML engineering is an experimental and iterative discipline, unlike traditional software engineering. Despite the negative perception of many failed experiments and models not being deployed, it is acceptable for them to not reach production."
    }
  ]
}
```

*Error*
```
{
  "status": "error",
  "message": "RetryError[<Future at 0x118d40700 state=finished raised AuthenticationError>]",
  "choices": []
}
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,6 @@
+APP_DATA_DIR=/Users/greg/Library/Application Support/com.tauri.dev
+PROJECT_NAME=project-x
 OPENAI_API_KEY=
-OPENAI_COMPLETE_MODEL=davinci
 OPENAI_CHAT_MODEL=gpt-3.5-turbo
 SIDECAR_LOG_DIR=/tmp
 SIDECAR_ENABLE_LOGGING=true
-
--- a/python/sidecar/chat.py
+++ b/python/sidecar/chat.py
@@ -17,6 +17,8 @@ openai.api_key = os.environ.get("OPENAI_API_KEY")
 def ask_question(request: ChatRequest):
     input_text = request.text
     n_choices = request.n_choices
+    temperature = request.temperature
+
     logger.info(f"Calling chat with the following parameters: {request.dict()}")
 
     storage = JsonStorage(filepath=settings.REFERENCES_JSON_PATH)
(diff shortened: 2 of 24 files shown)


.env.example
```diff
@@ -1,6 +1,6 @@
+APP_DATA_DIR=/Users/greg/Library/Application Support/com.tauri.dev
+PROJECT_NAME=project-x
 OPENAI_API_KEY=
-OPENAI_COMPLETE_MODEL=davinci
```
gjreda, line 2: OpenAI recently announced this is being deprecated and we do not use it.

python/sidecar/chat.py
```diff
@@ -39,6 +59,7 @@ def get_relevant_documents(self):
         docs = self.ranker.get_top_n(query=self.input_text, limit=5)
         return docs
 
+    @retry(stop=stop_after_attempt(3), wait=wait_fixed(1))
```
gjreda, line 62: I wasn't quite sure how we'd want to handle retries, but this felt like a fine starting point. In reality, I imagine if we need to spend more than a few second retrying (e.g. the user is rate limited or the service is down), we probably do not want to be retrying at all. Our user is still trying to write a document within Refstudio, even if the AI functionality is not working.

.env.example
cguedes: @gjreda should I then remove this from the configurations screen and fw to sidecar?

gjreda: @cguedes Yes, they'll be removing it from the API: https://openai.com/blog/gpt-4-api-general-availability

PR #309, opened 2023-07-18
Feature splashscreen

### LINKED ISSUE #308: Ensure settings are initialized before app starts
The application depends on settings to run, so we need to ensure that they are ready when the app first renders.

A solution is to use a [Tauri's splash screen](https://tauri.app/v1/guides/features/splashscreen/) while the app is loading.

cguedes:
This PR closes #308 by delaying app execution with a Tauri splash screen.

The splash screen is shown by default (main window is hidden by default) and then, when, the app is ready the command `close_splashscreen`) is executed and will hide splash screen and show the main window.

Note that the main window code still runs, while the app is not visible, so we have updated the root component to ensure the `<App>...</App>` component is only rendered when the app is ready.

Also, there is an initial white background in the splashscreen. Before rendering the content. There is a [pending issue reported](https://github.com/tauri-apps/tauri/issues/1564) to track that.

https://github.com/refstudio/refstudio/assets/174127/ec7d91e8-4b17-47ba-afe0-45322f541f06


src/main.tsx
```diff
@@ -1,20 +1,54 @@
...
+      interceptConsoleMessages(true, true, true);
+
+      notifyInfo('Application Startup');
+      await initSettings();
+      await ensureProjectFileStructure();
+      await invoke('close_splashscreen');
+      notifyInfo('Application Initialized');
+
+      setInitialized(true);
+    },
```
sehyod, line 44: I think you should use `isMounted`:
```typescript
async (isMounted) => {
      // Note: Intercepting INFO, WARN and ERROR console.*
      interceptConsoleMessages(true, true, true);

      notifyInfo('Application Startup');
      await initSettings();
      await ensureProjectFileStructure();
      await invoke('close_splashscreen');
      notifyInfo('Application Initialized');

      if (isMounted()) {
        setInitialized(true);
      }
    },
    noop,
    [],
  );
```
Even though it shouldn't be possible to unmount `AppStartup`, I think this is safer.

src/main.tsx
```diff
@@ -1,20 +1,54 @@
...
+    noop,
+    [setInitialized],
```
sehyod, line 46: You can remove it from the list of dependencies, setters don't change

sehyod (approved):

cguedes:
We can get rid of the white background by using a transparent splashscreen window with the `"macOSPrivateApi": true` config setting active. And configuring the background-color as transparent:

image

Note: Using the  `macOSPrivateApi` would prevent the app from being uploaded to the AppStore.

sehyod (approved):

PR #310, opened 2023-07-18
feat: allow multiple references in citation

### LINKED ISSUE #287: Allow for multiple references to be included in a single citation
In the [Markdown representation](https://quarto.org/docs/authoring/footnotes-and-citations.html#sec-citations) the keys will be separated by a semicolon.

sehyod:
Closes #287 

This adds a new `citation` node, that contains text and references.
The node can be created by typing `[` in the document.
`reference`s can now be inserted with the character `@`, but only in `citation` nodes
Inserting a `citation` automatically opens the reference picker
When at the end of a citation node, after a reference, typing anything will add `; ` before the added content

https://github.com/refstudio/refstudio/assets/58954208/dd9d55a9-be6b-4e0a-b7bd-468900224569

NB: References are now rendered with a React component, which means we can completely customise them, for example by having the name of the reference instead of the citation key, and making them hyperlinks that open the corresponding document. This also means that updating a reference in the reference table now updates it in the editor too.


hammer:
>  References are now rendered with a React component, which means we can completely customise them, for example by having the name of the reference instead of the citation key, and making them hyperlinks that open the corresponding document. This also means that updating a reference in the reference table now updates it in the editor too.

nice! Check out CSL for where we are going next … https://citationstyles.org/

src/features/textEditor/components/tipTapNodes/citation/citationNode.ts
```diff
@@ -0,0 +1,193 @@
...
+      ]);
+    },
+  },
+});
+
+/**
+ * Plugin that prevents the cursor from being directly after the opening and closing brackets of citation nodes
+ */
+const moveCursorPluginKey = new PluginKey('moveCursor');
+const moveCursorPlugin = new Plugin({
```
cguedes, line 122: not directly related to this (I think!) but can we make the cursor | visible inside the citation node?

src/features/textEditor/components/tipTapNodes/references/ReferencesList.tsx
```diff
@@ -9,10 +9,9 @@ import { usePopper } from 'react-popper';
 
 import { getReferencesAtom } from '../../../../../atoms/referencesState';
 import { cx } from '../../../../../lib/cx';
-import { ReferenceItem } from '../../../../../types/ReferenceItem';
 import { fuseOptions } from './config';
 
-export type ReferenceListProps = SuggestionProps<{ id: string; label: string }>;
+export type ReferenceListProps = SuggestionProps<{ id: string }>;
 
 export const ReferencesList = forwardRef((props: ReferenceListProps, ref) => {
```
cguedes, line 16: What do you think of renaming this as we already have one named like this one on the references sidebar.
Maybe `ReferencesListPopup`

cguedes (approved):
Great work 👏 . 

Just a note about which HTML element should we use for citations and some couple minor comments.

src/features/textEditor/components/tipTapEditorConfigs.ts
```diff
@@ -44,8 +45,9 @@ export const EDITOR_EXTENSIONS: Extensions = [
   CollapsibleBlockNode,
   CollapsibleBlockContentNode,
   CollapsibleBlockSummaryNode,
-  ReferenceNode,
-  SentenceCompletionExtension,
+  citationNode,
```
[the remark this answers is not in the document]
sehyod: I think I started using camel case to differentiate the TipTap extensions from React components, but we should keep names consistent. I will switch back to pascal case since that's the convention that TipTap uses.

src/features/textEditor/components/tipTapNodes/citation/citationNode.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  content: '(text | reference)*',
+  defining: true,
+  group: 'inline',
+  inline: true,
+  selectable: true,
+
+  parseHTML: () => [
+    {
+      tag: 'citation',
```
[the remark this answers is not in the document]
sehyod: That makes sense! I'm changing the reference node to render as a `cite` instead of `span`. Should I reset the style? Currently `cite` have `font-style: italic`

src/features/textEditor/components/tipTapNodes/citation/citationNode.ts
sehyod: I agree that is really annoying but I don't see any easy fix. The issue is that when there is a reference node followed by the end of the citation node, there is no text node in between (because empty text nodes are not valid), so the cursor doesn't show

cguedes (approved):

PR #313, opened 2023-07-19
138 search functionality new with json request and response

shauryr:
New branch with search functionality. 

- includes changes from the old branch https://github.com/refstudio/refstudio/pull/250
- Major changes made with https://github.com/refstudio/refstudio/pull/250#discussion_r1255895473 in mind. 

I was able to run the test and test the functionality via poetry. 
Please review. 
Screenshot 2023-07-19 at 1 28 16 PM


cguedes (changes requested):
It's working but we should have:

- improved Response type (for openAccessPdf)
- exception handling for S2 call

python/sidecar/typing.py
```diff
@@ -176,6 +176,27 @@ class ChatResponse(RefStudioModel):
...
+    venue: str | None = None
+    year: int | None = None
+    paperId: str | None = None
+    citationCount: int | None = None
+    openAccessPdf: Dict[str, str] | None = None
+    authors: list[str] | None = None
+    
+
+class SearchResponse(BaseModel):
+    status: str = "ok"
```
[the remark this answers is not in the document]
shauryr: sounds good! - https://github.com/refstudio/refstudio/pull/313/commits/1ccf725db5ab6d0abd41d3c0f1bf784d39fa97be

python/sidecar/search.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+
+        # Preprocess the query (e.g., lowercase, remove stopwords)
+        query = self.preprocess_query(query)
+
+        # Perform the search using Semantic Scholar's API
+        logger.info(f"Preprocessed query: {query}")
+
+        # Perform the search using Semantic Scholar's API
+        results = self.s2.search_paper(query, limit=limit, fields=returned_fields)
```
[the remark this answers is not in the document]
cguedes: It's great.

src/api/types.ts
```diff
@@ -169,3 +183,21 @@ export interface DeleteStatusResponse {
   status: ResponseStatus;
   message: string;
 }
+export interface SearchRequest {
+  query: string;
+  limit?: number;
+}
+export interface SearchResponse {
+  status: ResponseStatus;
```
cguedes, line 191: You forgot to add a `message` field here.

python/sidecar/search.py
```diff
@@ -0,0 +1,103 @@
...
+                year=getattr(item, "year", None),
+                paperId=getattr(item, "paperId", None),
+                citationCount=getattr(item, "citationCount", None),
+                openAccessPdf=openaccesspdf.get("url") if openaccesspdf else None,
+                authors=[author["name"] for author in getattr(item, "authors", [])],
+            )
+            results_list.append(result)
+
+        logger.info(f"First two papers from search results: {results_list[:2]}")
+        return SearchResponse(status="ok", results=results_list)
```
cguedes, line 96: You can something like this instead of your current return type:

* A `status` field of type `ResponseStatus`
* A `message` field of type `string` that is empty for successful responses and has the error message for exceptions
* A `results` field with `S2SearchResult` array (empty array for errors if you don't mind (like [here](https://github.com/refstudio/refstudio/blob/abab0993b4acd15bcf4d93fd0596d779afbdf05d/python/sidecar/chat.py#L36-L39) for chat)

```
    response = ChatResponse(
        status=typing.ResponseStatus.OK,
        message="",
        choices=[r.dict() for r in choices],
    )
```

cguedes (changes requested):
Update response types and code

src/api/types.ts
shauryr: thank you! fixed here https://github.com/refstudio/refstudio/pull/313/commits/1bb83ccd797ffd09f0b74dead9c4c17a8f9bbd74

python/sidecar/search.py
shauryr: thank you for sharing this. Does this look better - https://github.com/refstudio/refstudio/pull/313/commits/edc82ec0a2643790549d3fc8401fd87b6c452f45

cguedes: It does! Thanks.

python/sidecar/typing.py
```diff
@@ -176,6 +178,28 @@ class ChatResponse(RefStudioModel):
     choices: list[ChatResponseChoice]
 
 
+class SearchRequest(RefStudioModel):
+    query: str
+    limit: int = 10
+
+
+class S2SearchResult(BaseModel):
```
cguedes, line 186: Just noticed that you need to use `RefStudioModel` instead of `BaseModel` for a cleaner output.

shauryr: thank you for finding this. Fixed here - https://github.com/refstudio/refstudio/pull/313/commits/0e758a3fad237872a258a4450bf5eed644b49cb5

python/sidecar/typing.py
```diff
@@ -176,6 +178,28 @@ class ChatResponse(RefStudioModel):
...
+    paperId: str | None = None
+    citationCount: int | None = None
+    openAccessPdf: str | None = None
+    authors: list[str] | None = None
+
+
+class SearchResponse(RefStudioModel):
+    status: ResponseStatus
+    message: str
+    results: list[S2SearchResult] | None = None
```
cguedes, line 200: Just noticed now that that this `| None` is making the field optional in typescript. That is bad developer experience. We should follow the convention other responses used of not having the `None` and making sure that we return an empty list on error (even though I don't expect anyone would use that).

shauryr: fixed 👍  - https://github.com/refstudio/refstudio/pull/313/commits/1305958fbcdaa55deacb4cf2358e0e8875d0bf33

shauryr: updated - https://github.com/refstudio/refstudio/pull/313/commits/3c96863c75602f900ea299b2f3f81aaab32eb6a8

cguedes (approved):

PR #319, opened 2023-07-25
feat: Remove Open Editors section and move actions to context menu

cguedes:


src/application/App.tsx
```diff
@@ -81,6 +81,20 @@ function LeftSidePanelWrapper() {
     setPrimaryPane(selectedPane);
   };
 
+  // Configure keyboard shortcuts to open/close side panel
+  useEventListener('keydown', (e) => {
+    if (e.metaKey) {
+      switch (e.key.toLowerCase()) {
+        case 'b':
+          return handleSideBarClick(primaryPane);
```
sehyod, line 89: This is conflicts with cmd+B in the editor to make text bold: pressing cmd+B makes the text bold and toggles the sidebar. We need to prevent the event from bubbling out of the editor

src/components/TabPaneTabContextMenu.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+export function TabPaneTabContextMenu() {
+  const leftOpenEditorsData = useOpenEditorsDataForPane('LEFT');
+
+  const canMoveEditorToPane = (targetPane: PaneId) => (params: PredicateParams<TabPaneTabContextMenuProps>) => {
+    const isInLeft = leftOpenEditorsData.some((e) => e.id === params.props?.editorId);
+    if ((isInLeft && targetPane === 'RIGHT') || (!isInLeft && targetPane === 'LEFT')) {
+      return false;
+    }
+    return true;
```
sehyod, line 24: I think the function returns whether the editor **cannot** move to the given pane instead of whether it can.
You could simply do `return (isInLeft && targetPane === 'RIGHT') || (!isInLeft && targetPane === 'LEFT')`

src/components/TabPaneTabContextMenu.tsx
```diff
@@ -0,0 +1,85 @@
...
+    if (!editorId || !paneId || paneId === toPaneId) {
+      return;
+    }
+    emitEvent('refstudio://editors/move', { editor: { editorId, paneId }, toPaneId });
+  };
+
+  return (
+    <Menu animation={false} id={TABPANE_TAB_MENU_ID}>
+      <Item
+        disabled={canMoveEditorToPane('LEFT')}
```
sehyod, line 38: It should be `disabled={!canMoveEditorToPane('LEFT')}`

src/components/TabPaneTabContextMenu.tsx
```diff
@@ -0,0 +1,85 @@
...
+        }}
+      >
+        <div className="flex items-center gap-1">
+          <VscLayoutSidebarLeft />
+          Move Left
+        </div>
+      </Item>
+
+      <Item
+        disabled={canMoveEditorToPane('RIGHT')}
```
sehyod, line 50: Same here: `disabled={!canMoveEditorToPane('RIGHT')}`

src/components/TabPaneTabContextMenu.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+export const TABPANE_TAB_MENU_ID = 'TabPaneContextMenu';
+
+export interface TabPaneTabContextMenuProps {
+  editorId: EditorId;
+  paneId: PaneId;
+}
+
+export function TabPaneTabContextMenu() {
+  const leftOpenEditorsData = useOpenEditorsDataForPane('LEFT');
```
sehyod, line 17: You can greatly simplify the logic here since the props contain the `paneId`: there is no need to fetch `leftOpenEditorsData` to check whether the editor is in the left or right editor, you can just use the props attribute

src/events.ts
```diff
@@ -1,28 +1,38 @@
...
+  'refstudio://menu/file/close/all': undefined;
   'refstudio://menu/settings': undefined;
   'refstudio://menu/references/open': undefined;
   'refstudio://menu/references/upload': undefined;
   'refstudio://menu/references/export': undefined;
   'refstudio://menu/view/notifications': undefined;
+  'refstudio://menu/debug/console/clear': undefined;
+  // Editor actions
   'refstudio://editors/close': PaneEditorId;
+  'refstudio://editors/move': { editor: PaneEditorId; toPaneId: PaneId };
```
sehyod, line 20: Maybe you could rename it `fromPaneEditorId` to be more explicit? Or at least `paneEditorId`, in order not to get confused with the type of the attribute

src/wrappers/EventsListener.tsx
```diff
@@ -19,33 +24,30 @@ import {
...
   useListenEvent('refstudio://notifications/new', useCreateNotificationListener());
   useListenEvent('refstudio://notifications/clear', useClearNotificationsListener());
   // notifications popup
-  useListenEvent('refstudio://menu/view/notifications', useTauriViewNotificationMenuListener());
   useListenEvent('refstudio://notifications/popup/open', useShowNotificationsPopupListener());
   useListenEvent('refstudio://notifications/popup/close', useHideNotificationsPopupListener());
+  // View
+  useListenEvent('refstudio://menu/view/notifications', useTauriViewNotificationMenuListener());
+  // Debug
+  useListenEvent('refstudio://menu/debug/console/clear', useClearConsoleListener());
```
sehyod, line 50: Thank you for this clean-up 😄

src/application/App.tsx
cguedes: Nice catch. Maybe we just need to remove this shortcut.

src/application/commands/CommandPalette.tsx
```diff
@@ -96,6 +100,30 @@ export function MainCommandMenu({ index }: { index: number }) {
           text: 'New File',
           perform: () => emitEvent('refstudio://menu/file/new'),
         },
+        {
+          icon: <VscLayoutSidebarLeft />,
+          text: 'Move Left',
+          perform: () => {
+            if (activeEditorId) {
```
sehyod, line 107: Is there a way to disable or remove the item if there no active editor?

sehyod (approved):

src/application/commands/CommandPalette.tsx
cguedes: I don't know a way of doing that (that's easy). That's definitely something the library can improve in the future.

sehyod (approved):

sehyod (approved):

PR #322, opened 2023-07-26
feat: build a macOS Bundle

### LINKED ISSUE #311: Build a macOS bundle
https://tauri.app/v1/guides/building/macos

I think we need to put a `.app` file inside of a `.dmg` file but I'm not sure exactly. Either way it would be nice for people to install the app without having to clone the repo.

cguedes:
This PR prepares the code to create versioned releases of Ref Studio 🚀

- Instructions for the versioning rules and release were added to the root README file
- Update version to match the first version agreed on #311  (23.1.0)
- Add yarn build scripts to generate release files
- Fix the editor toolbar icons (replace images with react-icons).
- Add extra action in the editor toolbar to "Clear Formatting"

image
image

Note: After merging this into main we should tag main with 23.1.0 and create a release with the binaries.

Closes #311


src-tauri/Cargo.toml
```diff
@@ -1,10 +1,10 @@
 [package]
 name = "refstudio"
 version = "0.0.0"
-description = "A Tauri App"
+description = "An open source text editor optimized for writing that relies on references."
 authors = ["you"]
```
sehyod, line 5: Should we update `version` and `authors` here?

src-tauri/Cargo.toml
```diff
@@ -1,10 +1,10 @@
...
-license = ""
-repository = ""
+license = "MIT"
+repository = "https://github.com/refstudio/refstudio"
 edition = "2021"
```
sehyod, line 8: Shouldn't this be 2023?

hammer:
Is it easy to change the icon to be something specific to Ref Studio and not the Tauri icon? Also what's the best way to test out the `.dmg` file before we cut a release?

src-tauri/Cargo.toml
cguedes: No. that's a cargo configura! Go figure why the have their fields alongside with fields we can configure.

src-tauri/Cargo.toml
cguedes: AFAIK We don't need to update the version there, so I kept 0.0.0 in order to have a single place where we configure the version. 

I've updated the description bc it felt like a mandatory task. 

@hammer what should we put in the authors?

PS: Honestly, I don't know where these fields are being used currently. Most likely nowhere.

hammer:
I think we should just use the RS logo you put in the empty editor section on startup. I spoke with the design team today and they said explicitly they won't be doing logo design in this initial engagement.

cguedes:
> I think we should just use the RS logo you put in the empty editor section on startup. I spoke with the design team today and they said explicitly they won't be doing logo design in this initial engagement.

Ok, I will make one with the RS letters 🤞 😄

cguedes:
@hammer [this](https://github.com/refstudio/refstudio/pull/322/files#diff-f59990cf7bab634125197c888002439b08d313ab09ce165743fe08c5ba2963ed) is the new logo I've created and how it looks.

image

----

image

sehyod (approved):

PR #336, opened 2023-07-29
chore(backend): added logging to chunking

shauryr:
sometimes the backend is not able to chunk the full text from the pdfs because of file not found errors.

### DIFF
--- a/python/sidecar/shared.py
+++ b/python/sidecar/shared.py
@@ -9,7 +9,9 @@ import pypdf
 from sidecar import settings
 
 from .typing import Author, Chunk, Reference, TextCompletionChoice
+from .settings import logger
 
+logger = logger.getChild(__name__)
 
 def get_word_count(text: str) -> int:
     return len(text.strip().split(" "))
@@ -212,6 +214,7 @@ def chunk_reference(
     try:
         reader = pypdf.PdfReader(filepath)
     except FileNotFoundError:
+        logger.info(f"File not found: {filepath}")
         return []
 
     chunks = []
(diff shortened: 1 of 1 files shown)


gjreda:
@shauryr This LGTM, but do you have a sense of why this is happening? Was it just something that came up in your debugging?

I don't think it should have we happening in the app since the frontend knows all the file paths and passes them to the backend.

gjreda (approved):

shauryr:
> @shauryr This LGTM, but do you have a sense of why this is happening? Was it just something that came up in your debugging?
> 
> I don't think it should ever happen unless there is a bug on the front end, since the frontend knows all the file paths and passes them to the backend.

@gjreda when I ran the ingest command, the `chunks` variable was empty for me in `references.json` file. On closer inspection I found that for just chunking the code was not able to find the pdfs in the directory it was looking for. 
Then I found out that python is writing `.staging`, `.storage` and `.grobid` to a different folder than the UI. Python was writing these folder to a dir outside refstudio dir in my dev work folder while UI was doing it here `/Users/shaurya/Library/Application Support/studio.ref.desktop/project-x`. 

It maybe an issue because I have configured the project incorrectly. Am I doing something wrong here?

gjreda:
> Then I found out that python is writing .staging, .storage and .grobid to a different folder than the UI. Python was writing these folder to a dir outside refstudio dir in my dev work folder while UI was doing it here /Users/shaurya/Library/Application Support/studio.ref.desktop/project-x.

This sounds like it is related to your environment variables. The application passes the env variables on each call to the sidecar ([here](https://github.com/refstudio/refstudio/blob/main/src/api/sidecar.ts#L15-L28)), whereas if you are running just the python cli app, it will be using whatever env variables you have set.

Issue #140, opened 2023-06-12
Disambiguate, normalize, and link references against external databases


hammer:
We could use the S2 API or something like [Crossref](https://www.crossref.org/documentation/retrieve-metadata/simple-text-query/).

shauryr:
Adding [OpenAlex](https://docs.openalex.org/api-entities/works/search-works). This can be one of the external databases as well.

shauryr:
Once we have the references linked to an external database we can also fetch the full-text if it is available https://github.com/refstudio/refstudio/issues/141

PR #338, opened 2023-07-30
doi extraction for linking to external databases

shauryr:
Adds `doi` field from grobid extraction to the `references.json`. For most papers this can be used to link to other databases mentioned in https://github.com/refstudio/refstudio/issues/140


gjreda:
@shauryr Is this coming from grobid?

Can you also create a corresponding issue so that this is added to the UI as well? I imagine we'll want to make this field editable by the user in situations where we do not have the DOI, but the user would like to provide it.

gjreda (approved):

shauryr:
> 

@gjreda yes this is coming from grobid. 
Making the doi editable is a great suggestion! I will create an issue regarding this. Thank you!

PR #340, opened 2023-08-10
Link references to S2 based on DOI

shauryr:
https://github.com/refstudio/refstudio/issues/140
- The function `link_references` reads the `references.json` file and updates it with a `s2_paperId` field. 
- A reference is linked using the DOI field, which is extracted from the pdf. 
- This is the first step for linking to external databases. 

NEXT: If the DOI does not match, we have to match it with the title and other fields. This will be done in future PRs.


gjreda:
@shauryr Can you provide some more context here on the usage intent behind this feature? It'll help me in reviewing the PR.

For example, why add this as a method on `storage` rather than as part of the `ingest` process? What user action triggers this functionality?

Can you also add some tests please?

shauryr:
@gjreda Thank you for taking the time to review my PR and for raising these questions. I apologize for the lack of clarity in my initial submission. 

Once we have linked a reference to S2, we can verify and normalize the title, abstract and augment other information about a paper like field of study, citation count etc. 

In regard to your question on why this feature was added as a method on `storage` rather than as a part of the `ingest` process, the decision was based on initial considerations to maintain speed in the ingest process. However, in hindsight, I understand where you're coming from. Incorporating this feature into the ingest process would likely provide a more streamlined and effective solution.

I must admit that I haven't fully considered the exact user action that would trigger this feature. My approach was influenced by the structure of the `update_references` implementation, which this feature mirrored in some ways. 

Regarding your suggestion to add test cases, I appreciate your feedback and will work on incorporating that into the PR as soon as possible.

Once again, thank you for your valuable input.

PR #352, opened 2023-08-15
Update ingest to only use grobid for header parsing

### LINKED ISSUE #349: Update grobid ingest to only parse header data rather than full text
@shauryr called this out in #111

gjreda:
fixes #349


shauryr:
lgtm

shauryr (approved):

PR #356, opened 2023-08-16
Http stub fastapi

gjreda:


python/sidecar/chat.py
```diff
@@ -100,3 +101,4 @@ def ask_question(self, n_choices: int = 1, temperature: float = 0.7) -> dict:
         response = self.call_model(messages=messages, n_choices=n_choices, temperature=temperature)
         choices = self.prepare_choices_for_client(response=response)
         return choices
+        return choices
```
danvk, line 104: Looks like this last line got doubled up?

python/sidecar/search.py
```diff
@@ -106,3 +109,5 @@ def search_s2(request: SearchRequest):
     limit = request.limit
     searcher = Searcher()
     sys.stdout.write(searcher.search_func(query, limit=limit).json())
+    searcher = Searcher()
```
danvk, line 112: last two lines got doubled up here

python/sidecar/_typing.py
danvk, line 1: Does FastAPI force the `typing.py` -> `_typing.py` change? Is this just good form? We could also go with a different name like `types.py`.

python/main.py
```diff
@@ -1,8 +1,11 @@
 import inspect
 import json
 
-from sidecar import chat, cli, ingest, rewrite, storage, search
-from sidecar.typing import CliCommands
+from sidecar import chat, cli, ingest, rewrite, search, storage
+from sidecar._typing import CliCommands
+from sidecar.http import app
+
+app
```
danvk, line 8: what's going on here? is this a no-op?

python/sidecar/http.py
```diff
@@ -0,0 +1,50 @@
...
+
+app = FastAPI()
+
+
+@app.get("/")
+async def index():
+    return {"message": "Hello World"}
+
+
+@app.get("/rewrite")
```
danvk, line 15: We'll want all these routes to live under `/api` or even `/api/v0` to avoid collisions with static content handlers.

python/sidecar/http.py
```diff
@@ -0,0 +1,50 @@
...
+async def rewrite(
+    text: str,
+    manner: RewriteMannerType = "concise",
+    n_choices: int = 1,
+    temperature: float = 0.7
+) -> RewriteResponse:
+    request = RewriteRequest(
```
danvk, line 22: Is there any way to get FastAPI to give you the request body as a `RewriteRequest` object?

danvk (commented):
Can you share some examples of what calling this HTTP API looks like, ideally from the command line using `curl`?

python/sidecar/_typing.py
gjreda: The root of this was that one of the underlying libraries used by fastapi and litestar was doing a `import typing` which was causing circular import issues with our `typing.py` (I've run into similar issues when named `types.py` too).

Though I'm playing around with this a bit more on a different branch now and I don't think it's an issue anymore 🤷

python/main.py
gjreda: Effectively. Really it was just me being lazy and wanting to simply run `uvicorn main:app` to run the server. After a couple tries of unsuccessfully getting it to run from within `http.py` I just decided to import it here for the sake of a prototype.

python/sidecar/chat.py
gjreda: Yea, there are a couple more spots too -- something was up with my VS Code + vim plugin

python/sidecar/http.py
gjreda: Yep! I've got that working on a different branch now.

PR #359, opened 2023-08-16
Add http endpoints for sidecar functionality

gjreda:
### Running this PR
```bash
cd refstudio/python

poetry run uvicorn web:api --reload
```
You should then be able to visit http://127.0.0.1:8000/api/v0/docs in your browser and test various endpoints. Not all of them have been implemented yet, and some need some slight changes due to the way the current sidecar works (i.e. since we communicate via stdout, maybe of the existing sidecar functions do not `return` anything ... which needs to happen for the API).

### Example usage
```
curl -s -X 'POST' \
  'http://127.0.0.1:8000/api/sidecar/ai/rewrite' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "text": "Chicago is the most populous city in the U.S. state of Illinois and the third-most populous in the United States after New York City and Los Angeles. With a population of 2,746,388 in the 2020 census, it is also the most populous city in the Midwest. As the seat of Cook County (the second-most populous U.S. county), the city is the center of the Chicago metropolitan area, one of the largest in the world.",
  "manner": "concise",
  "n_choices": 1,
  "temperature": 0.7
}' | jq

{
  "status": "ok",
  "message": "",
  "choices": [
    {
      "index": 0,
      "text": "Chicago is the most populous city in Illinois and the third-most populous in the United States, after New York City and Los Angeles. It has a population of 2,746,388 according to the 2020 census, making it the most populous city in the Midwest. It is also the center of the Chicago metropolitan area, which is one of the largest in the world and is located in Cook County, the second-most populous county in the US."
    }
  ]
}
```


python/sidecar/http.py
```diff
@@ -0,0 +1,85 @@
...
+    RewriteResponse,
+    SearchRequest,
+    SearchResponse,
+    TextCompletionRequest,
+    TextCompletionResponse,
+    UpdateStatusResponse,
+)
+
+app = FastAPI()
+v0 = FastAPI()
```
danvk, line 23: `v0` -> `api`?

python/sidecar/http.py
```diff
@@ -0,0 +1,85 @@
...
+@v0.post("/ingest/upload")
+async def upload() -> IngestResponse:
+    # response = sidecar.ingest.run_ingest()
+    raise NotImplementedError("Upload is not yet implemented")
+
+
+@v0.get("/ingest/status")
+async def get_statuses() -> IngestStatusResponse:
+    response = sidecar.ingest.get_statuses()
+    return response
```
danvk, line 40: any reason to keep the `response` variable here instead of `return sidecar.ingest.get_statuses()`? (also below)

python/sidecar/http.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+@v0.get("/ingest/references")
+async def get_references(req: IngestRequest) -> IngestResponse:
+    response = sidecar.ingest.get_references(req)
+    return response
+
+
+@v0.post("/rewrite")
+async def rewrite(req: RewriteRequest) -> RewriteResponse:
```
danvk, line 50: Does FastAPI validate request bodies based on these type annotations?

python/main.py
```diff
@@ -1,9 +1,11 @@
 import inspect
 import json
 
-from sidecar import chat, cli, ingest, rewrite, storage, search
+from sidecar import chat, cli, http, ingest, rewrite, search, storage
 from sidecar.typing import CliCommands
 
+app = http.app
```
danvk, line 7: Can we keep these separate? The web server isn't necessary for running the CLI.

python/sidecar/http.py
danvk, line 1: This probably doesn't belong in the `sidecar` dir since it doesn't (shouldn't) run as part of the sidecar.

danvk (dismissed):
Looks good! Two high-level bits of feedback:

1. It would be nice to have some tests
2. Does this do validation of request bodies for us?

We'll want to get these endpoints into our codegen setup but that can be in a follow-up PR.

Also I'm seeing this error when I try to use the `GET /ingest/references` endpoint:

```
TypeError: Failed to execute 'fetch' on 'Window': Request with GET/HEAD method cannot have body.
```

image

cguedes (changes requested):
Agree with Dan's comments.

We should have a distinct entry point for the HTTP server (from the cli/sidecar).

I see the backend for HTTP to translate http parameters (body, querystring) and call the internal functions that return a value. Then, that value is sent back to the client serialized as JSON.

For the CLI application interface, I see the app receiving requests in JSON (...Request) and also call the internal function that **return a value**. Then is the responsibility of the CLI to translate that to a `sys.stdout.write(response.json())` reply.

The HTTP endpoints should also be adjusted to:
- use the first segment to scope the different backend scopes:
       - references (ingestion, status, edit, delete, ...), 
       - ai (completion, chat)
       - search (s2)
 - adopt the best HTTP method
       - `GET` to access a resource or query information
            - I think we should use this method for text completion and for s2 search)
       - `DELETE` to delete a resource (no need to repeat delete in the URL, we should have an URL that identify a resource)
       - `PATCH` to partially update a reference

Alternatively, if we want to mimic the sidecar request/reply method of sending a JSON payload and receiving a JSON payload, we should only use `POST` to a `api/v0/sidecar/...` endpoint.

python/sidecar/http.py
gjreda: Nah, just stylistic and consistency with the other functions

python/sidecar/http.py
gjreda: yes, here's an example:

```bash
curl -X 'POST' \
  'http://127.0.0.1:8000/api/v0/completion' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "n_choices": 1,
  "temperature": 0.7,
  "max_tokens": 512,
  "title": "string",
  "abstract": "string"
}'

{"detail":[{"loc":["body","text"],"msg":"field required","type":"value_error.missing"}]}
```

python/sidecar/http.py
gjreda: do you think these endpoints should live under `/api/` or `/api/v0`? I'm indifferent.

if the latter, if feels like it makes sense to scope them to `v0` or `api_v0`

gjreda:
> For the CLI application interface, I see the app receiving requests in JSON (...Request) and also call the internal function that return a value. Then is the responsibility of the CLI to translate that to a sys.stdout.write(response.json()) reply.

@cguedes I agree with this approach, though it's a fairly sizable refactor (mainly the tests) that I think would be distracting in this PR. I'll add it as a follow up.

python/main.py
```diff
@@ -1,7 +1,7 @@
 import inspect
 import json
 
-from sidecar import chat, cli, ingest, rewrite, storage, search
+from sidecar import chat, cli, ingest, rewrite, search, storage
```
danvk, line 4: Not necessary for this PR, but is there a linter/formatter that's not running as part of CI?

danvk (approved):

python/main.py
gjreda: Honestly, this is something my vs-code only does for refstudio and I haven't been able to figure out why

gjreda: I had thought it might have been due to prettier integration, but seems not

danvk (approved):

cguedes (approved):

Issue #347, opened 2023-08-14
Allow Ref Studio to run locally as a web app


hammer:
Order of operations here, at a high level, maybe:
- Refactor Python sidecar to run as a service or set of services
- RS Desktop frontend talks to sidecar service(s) over websockets
- RS frontend runs within web server

I'm mostly interested in local deployment for now, we can file a separate issue to host it somewhere once we get local deployment working.

hammer:
Reminder that [#17](https://github.com/refstudio/refstudio/issues/17) had some discussion of sidecar over websockets

cguedes:
This is the current version of the app architecture changes to support desktop and web version of RefStudio. 

![image](https://github.com/refstudio/refstudio/assets/174127/c97553bd-00df-45e8-a37c-e5979e8851bf)


**Sidecar**

- [ ] HTTP server for the python backend commands
     - ex: `POST /ingest/upload { body }`, `GET /ingest/references`, `GET /ingest/status`, `POST chat/...`, `GET /search/...`
- [ ] Update ingestion API to receive the PDF files instead of PDF files path
     - we should support both methods (PDF files uploaded and the current path) for now
     - for the PDF files upload version, we need an additional parameter with the project name (or id) to support multiple projects. That would be useful to scope the multiple `references.json` files in the backend server supporting filesystem
- [ ] Frontend `sidecar.ts` facade that calls Tauri or HTTP backend APIs
     - Maybe we can (should) rename sidecar to something more agnostic
- [ ] Reimplement flow for PDF uploads to work with the new HTTP backend API

**Application Launcher**

- [ ] Run the application inside a web server
    - We can start by running `yarn dev` to run vite server (change initialised to true in the AppStartup to test)
    - We might remove the splash-screen to simplify code and implement a "html/react/native" splash-screen
    - We might use Vite `import....` to conditionally build the app and have code splitting using System.Import
- [ ] Adjust SettingsManager to work in web
    - In web, we need to not read env variables (or we can have a Tauri facade for web that just "noop" env vars)
    - In web, we can't load/save settings to a file

**Filesystem**

- [ ] HTTP server (python) for a virtual filesystem
- [ ] Facade that supports remote filesystem to work in the web
- [ ] Facade that supports local filesystem to work in the desktop Tauri

**Menu & Events**

- [ ] Explore the support for dynamically create (and update) Tauri menu from the TS code
- [ ] Menu component for the web version
- [ ] Facade to support events in the web version


[...]

PR #368, opened 2023-08-17
Codegen menus (POC)

danvk:
Our dream was to define the menu bar on the TypeScript side and have it either get pushed into the system menu bar (for the desktop app) or get rendered in the DOM (for the web UI).

That dream doesn't seem possible at the moment, so instead we have codegen! This PR creates `menu.json` as a source of truth and generates `menu.rs` and `menu.tsx` from it. The `menu.rs` file is almost byte-for-byte identical with the hand-generated one. `menu.tsx` doesn't do anything at the moment, but it's what we'd want to render a menu bar in the web version of RefStudio (#347).

Looking at `menu.tsx`, I'm actually wondering whether this is a good idea after all. A pretty large fraction of the menu items are system-related ones ("Hide Others", "Maximize", "Select All") that might not make sense to show in a DOM UI since they're already part of the browser's system menu bar.


.vscode/settings.json
```diff
@@ -28,6 +28,7 @@
   },
   "[rust]": {
     "editor.defaultFormatter": "rust-lang.rust-analyzer",
-    "editor.formatOnSave": true
+    "editor.formatOnSave": true,
+    "editor.inlayHints.enabled": "offUnlessPressed"
```
cguedes, line 32: @danvk thanks for this setting!

danvk: Split this off into #403

danvk:
I'm going to close this out.

Another wrinkle from #399 is that keyboard shortcuts might need to be different between Desktop. If we want a menu bar, then two separate implementations might really be the way to go. And, per #384, we may not even want a menu bar at all.

PR #382, opened 2023-08-18
feat: web call sidecar http

### LINKED ISSUE #380: Frontend sidecar.ts facade that calls Tauri or HTTP backend APIs
Related to #347 

After #358 and #359 we can update the codebase to call the HTTP backend when deployed in the browser.

cguedes:
This PR fixes #380 that allows the web application to call the sidecar commands via HTTP interface.
Also adds utility startup scripts and updates README.md

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,6 +13,11 @@
   "[typescript]": {
     "editor.formatOnSave": true
   },
+  "[python]": {
+    "editor.formatOnSave": false,
+    "editor.defaultFormatter": "ms-python.python",
+    "editor.codeActionsOnSave": []
+  },
   "editor.codeActionsOnSave": [
     // Run ESLint fixers _before_ prettier to ensure the user doesn't need to do multiple saves
     "source.organizeImports",
--- a/python/sidecar/http.py
+++ b/python/sidecar/http.py
@@ -1,13 +1,29 @@
 from dotenv import load_dotenv
 from fastapi import FastAPI
-from sidecar import chat, ingest, rewrite, search, storage
-from sidecar.typing import (ChatRequest, ChatResponse, DeleteRequest,
-                            DeleteStatusResponse, IngestRequest,
-                            IngestResponse, IngestStatusResponse,
-                            ReferenceUpdate, RewriteRequest, RewriteResponse,
-                            SearchRequest, SearchResponse,
-                            TextCompletionRequest, TextCompletionResponse,
-                            UpdateStatusResponse)
+from sidecar import (
+    chat as chat_module,
+    ingest as ingest_module,
+    rewrite as rewrite_module,
+    search as search_module,
+    storage
+)
+from sidecar.typing import (
+    ChatRequest,
+    ChatResponse,
+    DeleteRequest,
+    DeleteStatusResponse,
+    IngestRequest,
+    IngestResponse,
+    IngestStatusResponse,
+    ReferenceUpdate,
+    RewriteRequest,
+    RewriteResponse,
+    SearchRequest,
+    SearchResponse,
+    TextCompletionRequest,
+    TextCompletionResponse,
+    UpdateStatusResponse,
+)
 
 load_dotenv()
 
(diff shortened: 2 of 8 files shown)


python/sidecar/http.py
```diff
@@ -1,24 +1,13 @@
...
-    ReferenceUpdate,
-    RewriteRequest,
-    RewriteResponse,
-    SearchRequest,
-    SearchResponse,
-    TextCompletionRequest,
-    TextCompletionResponse,
-    UpdateStatusResponse,
-)
+from sidecar import chat, ingest, rewrite, search, storage
```
cguedes, line 3: @gjreda please review this file changes bc I've had some issues calling the API endpoints in my local env!

Commit with the change (and more details) can be found [here](https://github.com/refstudio/refstudio/pull/382/commits/41df52f4817174d59075200b22069e085af3b823).

python/sidecar/http.py
```diff
@@ -1,24 +1,13 @@
...
+from sidecar.typing import (ChatRequest, ChatResponse, DeleteRequest,
+                            DeleteStatusResponse, IngestRequest,
+                            IngestResponse, IngestStatusResponse,
+                            ReferenceUpdate, RewriteRequest, RewriteResponse,
+                            SearchRequest, SearchResponse,
+                            TextCompletionRequest, TextCompletionResponse,
+                            UpdateStatusResponse)
```
gjreda, line 10: Can we leave the formatting here? I suspect prettier is doing something to Python files but I haven't been able to figure out how to stop it.

python/sidecar/http.py
```diff
@@ -31,54 +20,54 @@ async def sidecar_index():
...
-    # response = sidecar.ingest.run_ingest()
+async def http_ingest() -> IngestResponse:
+    # response = ingest.run_ingest()
     raise NotImplementedError("Ingest is not yet implemented")
 
 
 @sidecar_api.get("/ingest_status")
-async def get_statuses() -> IngestStatusResponse:
-    response = sidecar.ingest.get_statuses()
+async def http_get_statuses() -> IngestStatusResponse:
```
gjreda, line 29: Why prefix these with `http`?

python/sidecar/http.py
```diff
@@ -31,54 +20,54 @@ async def sidecar_index():
 
 
 @sidecar_api.post("/ingest")
-async def upload() -> IngestResponse:
-    # response = sidecar.ingest.run_ingest()
+async def http_ingest() -> IngestResponse:
```
gjreda, line 23: Why rename this one?

python/sidecar/http.py
cguedes: I've had to prefix the `chat` and then I’ve used the same rule for all the other methods. A bit of OCD here I must admit :-).

Also, this is something I do in my projects when implementing HTTP apis, where I normally use `*Handler` (ex: `chatHandler`). I can use that convention if you want.

gjreda: I would prefer to not prefix them all with http_ since they are already namespaced by nature of being in the http module. If there is a name collision for chat I would just do either:
```
from sidecar import chat as chat_

or 

from sidecar import chat as chat_module
```

cguedes: Ok. I will make that change instead.

python/sidecar/http.py
cguedes: To match the sidecar command name and url (/ingest) like you have for most of the commands.
Also, IMO we won’t have upload available via the sidecar api. Right? 
And I think that you will implement a different file/url/code for the upload feature.

I will also revert this change.

gjreda: Given our Slack huddle, I'm ok if we keep this one named `ingest`

.vscode/settings.json
```diff
@@ -13,6 +13,11 @@
   "[typescript]": {
     "editor.formatOnSave": true
   },
+  "[python]": {
+    "editor.formatOnSave": false,
```
cguedes, line 17: @gjreda added this configurations to prevent VSCode from automatically format the python code with prettier.
Let me know what is the code formatter we should use and I can configure it instead of "ms-python.python".

PR #383, opened 2023-08-18
feat: update sidebar design

### LINKED ISSUE #371: Design for PrimarySideBar (plus PanelWrapper and PanelSection)

sehyod:
This adds a generic Sidebar component, uses it for the left sidebar, and replaces the right panel logic with a second sidebar on the right
This also fixes the background of the splashscreen to be the same as the one of the welcome screen, instead of white

Screenshot 2023-08-22 at 14 43 06

NB: This PR does **not** update the sidebar panels (like the references panel, the file explorer panel, etc.)

Closes #371


tailwind.config.js
```diff
@@ -21,6 +21,7 @@ export default {
         'grayscale-10': 'hsl(var(--grayscale-10) / <alpha-value>)',
         'grayscale-20': 'hsl(var(--grayscale-20) / <alpha-value>)',
         'grayscale-30': 'hsl(var(--grayscale-30) / <alpha-value>)',
+        'grayscale-40': 'rgb(219 223 231 / <alpha-value>)',
```
sehyod, line 24: I am defining this one differently because having the rgb value here is easier and we don't need the CSS variables.
I am not updating the other variables because we will remove them and replace them with variables with proper names shortly

src/application/App.tsx
```diff
@@ -127,11 +131,28 @@ function LeftSidePanelWrapper() {
...
+            onClick: openSettings,
+          },
+        ]}
+        items={[
+          { pane: 'Explorer', Icon: <FilesIcon /> },
+          { pane: 'References', Icon: <ReferencesIcon /> },
+        ]}
+        onItemClick={handleSideBarClick}
       />
+      {!primaryPaneCollapsed && <div className="h-full w-[1px]" style={{ backgroundColor: '#eff1f4' }} />}
```
cguedes, line 155: Why do we need this? Also, with that hard-coded color.

src/application/App.tsx
```diff
@@ -141,44 +162,47 @@ function LeftSidePanelWrapper() {
   );
 }
 
-function RightPanelWrapper() {
-  const [closed, setClosed] = React.useState(false);
-  const panelRef = React.useRef<ImperativePanelHandle>(null);
+type SecondarySideBarPane = 'Rewriter' | 'Chatbot';
+function RightSidePanelWrapper() {
```
cguedes, line 166: What do you think of moving the Left and Right side panel wrappers to own files (or both in same file)?

cguedes (changes requested):

src/application/App.tsx
sehyod: Oops, that comes from my initial tests, when I hadn't created the tailwind variables, I forgot to update it.

src/application/App.tsx
sehyod: Yes, that makes sense, and will make the code more legible

src/application/App.tsx
cguedes: Ok. I see that this is a vertical line only visible when the primaryPanel is visible.

cguedes (approved):

PR #388, opened 2023-08-21
feat: menu shortcuts for web

### LINKED ISSUE #387: Menu shortcut action available on the web version
Part of work for https://github.com/refstudio/refstudio/issues/347

We should have menu actions, and shortcuts, available in the web. They should also be in the command palette.

cguedes:
Closes #387 and related to #347 

This PR adds menu actions to the command palette and shortcut handlers for menu actions.
Also stop event propagation for existing commands to prevent tab switch in the web for example (cmd+1, ...).

image


src/application/App.tsx
```diff
@@ -108,9 +110,15 @@ function LeftSidePanelWrapper() {
...
+          handleSideBarClick('Explorer');
+          e.preventDefault();
+          e.stopPropagation();
+          break;
         case '2':
-          return handleSideBarClick('References');
+          handleSideBarClick('References');
+          e.preventDefault();
+          e.stopPropagation();
+          break;
```
sehyod, line 121: nit: you could prevent default and stop propagation before the switch case so you don't have to write it multiple times

src/application/WebMenuShortcuts.tsx
```diff
@@ -0,0 +1,57 @@
...
+ *    meta + up + 1..9 on the other hand focuses the corresponding tab of the active window and
+ *    also cannot be overridden.
+ */
+export function WebMenuShortcuts() {
+  // RefStudio
+  useMenuEventHotKeyForWeb(['meta+,'], 'refstudio://menu/settings');
+  // File
+  useMenuEventHotKeyForWeb(['ctrl+n'], 'refstudio://menu/file/new');
+  useMenuEventHotKeyForWeb(['ctrl+s', 'meta+s'], 'refstudio://menu/file/save');
+  useMenuEventHotKeyForWeb(['ctrl+w'], 'refstudio://menu/file/close');
```
sehyod, line 30: I think these wouldn't work on Windows, since Windows use ctrl for this kind of shortcuts

src/application/App.tsx
cguedes: No, I don't want to do that for other shortcuts than CMD+1 and CMD+2.

src/application/WebMenuShortcuts.tsx
```diff
@@ -0,0 +1,57 @@
...
+  // References
+  useMenuEventHotKeyForWeb(['meta+r'], 'refstudio://menu/references/open');
```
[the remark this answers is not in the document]
cguedes: That's a good point. This shortcut is also tricky because of that. I've found (for chrome at least) that it works with your code (preventing page refresh) for the first time you press. But if you press the shortcut again, it would refresh the page. Smart implementation!

src/application/WebMenuShortcuts.tsx
cguedes: That's a good point. I feel that, for the web, we might drop (not use) most of these shortcuts for window/file management.

src/hooks/useRefStudioHotkeys.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Configure keyboard shortcut with some arbitrary handler.
+ * Note that this will run both for desktop and web
+ *
+ * @param keys keyboard shortcut configurations
+ * @param handler handler executed on shortcut press
+ * @param onlyForWeb if true, keyboard shortcut is only configured for the web
+ */
+export function useRefStudioHotkeys(keys: Keys, handler: () => void, onlyForWeb = false) {
```
sehyod, line 21: I don't think you ever use `onlyForWeb`

src/hooks/useRefStudioHotkeys.ts
```diff
@@ -0,0 +1,59 @@
...
+/**
+ *
+ * @param keys keyboard shortcut configurations
+ * @param onlyForWeb if true, keyboard shortcut is only configured for the web
+ * @param eventName ref studio event name
+ * @param eventPayload ref studio event payload
+ */
+export function useRefStudioEventHotkeys<Event extends RefStudioEventName>(
+  keys: Keys,
+  onlyForWeb: boolean,
```
sehyod, line 41: Same here, we can probably remove the parameter

sehyod (approved):

PR #389, opened 2023-08-21
feat: Add filesystem and projects API

gjreda:
fixes #365
__________

This PR adds APIs for both managing files and projects.

### Running the server
```bash
cd python; poetry run uvicorn web:api --reload
```

### Create a project
```bash
curl -X 'POST' \
  'http://127.0.0.1:8000/api/projects/' \
  -H 'accept: application/json' \
  -d '' -s | jq

{
  "63f40666-6aae-49c6-b375-be09ca3dd5bc": "/tmp/web-storage-url/user1/63f40666-6aae-49c6-b375-be09ca3dd5bc"
}
```

### List projects
```bash
curl -X 'GET' \
  'http://127.0.0.1:8000/api/projects/' \
  -H 'accept: application/json' -s | jq

{
  "project1": "/tmp/web-storage-url/user1/project1",
  "project2": "/private/var/folders/4v/0m97cs6j7v53r6m3mwg340380000gn/T/pytest-of-greg/pytest-389/test_update_project_path_stora1/user1/project2",
  "d94ca1db-f251-48c3-aafe-67be329b6006": "/tmp/web-storage-url/user1/d94ca1db-f251-48c3-aafe-67be329b6006",
  "63f40666-6aae-49c6-b375-be09ca3dd5bc": "/tmp/web-storage-url/user1/63f40666-6aae-49c6-b375-be09ca3dd5bc"
}
```

### Get project details
```bash
curl -X 'GET' \
  'http://127.0.0.1:8000/api/projects/63f40666-6aae-49c6-b375-be09ca3dd5bc' \
  -H 'accept: application/json' -s | jq

{
  "project_id": "63f40666-6aae-49c6-b375-be09ca3dd5bc",
  "project_path": "/tmp/web-storage-url/user1/63f40666-6aae-49c6-b375-be09ca3dd5bc",
  "filepaths": []
}
```

### Delete a project
```bash
curl -X 'DELETE' \
  'http://127.0.0.1:8000/api/projects/63f40666-6aae-49c6-b375-be09ca3dd5bc' \
  -H 'accept: application/json'

{
  "status": "success",
  "message": "Project deleted",
  "project_id": "63f40666-6aae-49c6-b375-be09ca3dd5bc"
}
```

### List project files
```bash
curl -X 'GET' \
  'http://127.0.0.1:8000/api/projects/63f40666-6aae-49c6-b375-be09ca3dd5bc/files' \
  -H 'accept: application/json'
```
_______

### Create a file (this will respect directory segments)
[...]

PR #390, opened 2023-08-22
feat: Add python tests script and run python tests in pre-push

### LINKED ISSUE #391: bug: python tests don't run in the git push hook
We should have python tests running in the git push hook.

I've just missed the following failing python test when merging #389.

image

cguedes:
@gjreda I've just noticed that the #389 is failing in the CI after merge agains main!

This PR adds an utility to run python tests from the project root using yarn `yarn python:tests` and updates the git push hook to also run python tests.

@gjreda please fix the error reported in #391 (I don't know how 😅).
_____
fixes #391

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -3,5 +3,6 @@
 
 yarn check
 yarn test
+yarn python:test
 
 yarn codegen
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "ts:check": "tsc --noEmit",
     "preview": "vite preview",
     "python": "bash scripts/remove_binary.sh && poetry run pyinstaller --noconfirm --distpath src-tauri/bin/python python/main.py && bash scripts/move_binary.sh",
+    "python:test": "poetry run pytest python/tests",
     "test": "vitest run",
     "test:watch": "vitest --watch",
     "test:watch:ui": "vitest --watch --ui",
--- a/python/sidecar/projects.py
+++ b/python/sidecar/projects.py
@@ -4,27 +4,33 @@ import shutil
 
 from sidecar import settings
 
-"""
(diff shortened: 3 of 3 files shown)


gjreda:
@cguedes The tests started to fail when you pushed this commit to the PR: https://github.com/refstudio/refstudio/pull/389/commits/f0266033ed058e459b68b7b0fbf87e54140b8f58

I'll push a fix, but they were failing before merging into main. The root cause is that you [changed the filename](https://github.com/refstudio/refstudio/pull/389/commits/f0266033ed058e459b68b7b0fbf87e54140b8f58#diff-aecd2e6bb1e4f5c776d51ddd0be8303d9b662d1c71f40648f0106447c6c513fcR14) for projects storage, but did not update the tests.

gjreda:
I suspect the bigger bug here is in your merge process: since we have historically ignored when codecov/patch is failing, I'm going to guess you thought that was the cause of the ❌ on the PR, and not that actual tests were failing.

cguedes:
> I know I've been guilty of this in the past too, so we should probably figure out how to make codecov/patch work as we'd like or simply get rid of it, since we do not trust it.

You are right. I thought it was again the codecov/patch issue.
We need to fix our merge process and get rid of codecov/patch. @sehyod what do you think of this?

gjreda:
fixes #391

sehyod:
> > I know I've been guilty of this in the past too, so we should probably figure out how to make codecov/patch work as we'd like or simply get rid of it, since we do not trust it.
> 
> You are right. I thought it was again the codecov/patch issue. We need to fix our merge process and get rid of codecov/patch. @sehyod what do you think of this?

Yes, I 100% agree! codecov/patch is not super useful anyway, the most important is codecov/project

PR #392, opened 2023-08-22
Export references

### LINKED ISSUE #345: Export references as .bib
One thought: export on full references list vs. export only those references cited in a document. Also, do we need to serialize references cited in a document when we save the document, or just keep the cite key?

sehyod:
This adds the function to exports the entire list of references as a .bib file. Clicking on the button opens a dialog to choose where the file should be saved (defaulting to the `exports` folder of the project)
When no reference has been ingested, the button is disabled

Closes #345


src/io/filesystem.ts
```diff
@@ -335,6 +337,31 @@ export async function renameFile(relativePath: string, newName: string): RenameF
 }
 type RenameFileResult = Promise<{ success: false } | { success: true; newPath: string }>;
 
+export async function exportReferences(references: ReferenceItem[]) {
```
cguedes, line 340: nit: you need to add this to the  `__mocks__/filesystem.ts`

cguedes (approved):

PR #395, opened 2023-08-22
feat: Get file ingestion running through http

### LINKED ISSUE #366: Get file ingestion running through HTTP
Part of the work for #347 

Will need to add file upload functionality as part of this.

gjreda:
fixes #366 

This started as getting ingestion running through HTTP, but led to some new APIs so that we can move away from the sidecar CLI.

This PR adds the idea of a `project_name` to the `projects` core logic, removes some of the legacy `/sidecar` APIs and also adds the following:

## References
- List references => `GET /api/references/{project_id}`
- Get a reference => `GET /api/references/{project_id}/{reference_id}`
- Ingest references => `POST /api/references/{project_id}`
- Update a reference => `PUT /api/references/{project_id}/{reference_id}`
- Delete a reference => `DELETE /api/references/{project_id}/{reference_id}`
- Bulk delete references => `POST /api/references/{project_id}/bulk_delete`

## AI
- Chat => `POST /api/ai/{project_id}/chat`


python/sidecar/http.py
```diff
@@ -35,16 +35,22 @@
 
 # Sidecar API
 # -----------
-@sidecar_api.post("/ingest")
+@sidecar_api.get("/")
+async def http_index():
+    return {"message": "Hello World"}
+
+
+@sidecar_api.post("/ingest/{project_id}")
```
gjreda, line 43: @cguedes Let me know if you disagree with this approach

python/sidecar/http.py
```diff
@@ -35,16 +35,22 @@
...
     project_path = projects.get_project_path(user_id, project_id)
-    req = IngestRequest(pdf_directory=project_path)
+    uploads_dir = project_path / "uploads"
+    req = IngestRequest(pdf_directory=str(uploads_dir))
     response = ingest.run_ingest(req)
     return response
 
 
-@sidecar_api.get("/ingest_status")
+@sidecar_api.get("/ingest_status/")
```
gjreda, line 53: Maybe we made this take {project_id} as well?

gjreda:
@cguedes This gets ingest working, but I'd appreciate your feedback on the other `ingest` related endpoints -- `/ingest`, `/ingest_status`, and `/ingest_references`. I feel like we should change those endpoints to accept a `project_id`, but I also don't want to break other desktop-related things for y'all on the client side.

python/sidecar/http.py
cguedes: @gjreda I do 😅 

For the sidecar commands I think that we should match the existing architecture of sending via ENV var (like we do for project path) given that most (if not all) commands would need that. 

For the ingestion (ingest command) I would also, at least for now, make the `uploads` folder fixed instead of configurable via request payload, and therefore drop the `pdf_directory` in the `IngestRequest`.

Additionally, to match the http interface, we should also only depend on the project_id and create additional sidecar commands for create and open a project.

We can have a call as soon as you start working today to align the implementation of the backend and frontend.

CC @danvk maybe also with you.

python/sidecar/http.py
cguedes: Maybe not. check the other comment.

python/sidecar/http.py
gjreda: Thinking about this more, my concern with sending things [via env vars like you do here](https://github.com/refstudio/refstudio/pull/393/files#diff-0d789014de4ae1d8a4ec1e870aa5e951f6a25db3fdc3732cb341948e0fbee147R46) is that the API is no longer obvious. It becomes harder to reason about what is truly required for a given request and makes the backend code a bit more mysterious. I also don't think we'll need things like the API key on each request, since it should be set via the settings API.

gjreda: I'd rather we get rid of the env variables on the cli side and make things explicit rather than relying on env-like variables on the http side.

cguedes: Ok. I don’t mind (actually prefer) to have that parameter being part of the payload every time we need it.

danvk: Did you guys come to a resolution here? I always like using the GitHub API as a reference on how to structure things. They include things like org and repo in the request path rather than in headers. Headers and query parameters are used for things that don't refer to the resource itself, like authorization and response format.

cguedes: @danvk yes, we did. We will have the {project_id} in the new apis for references:

- `GET `api/references/{project_id}` (get all references}
- `POST api/references/{project_id}` (ingest what is in /uploads)
- `GET api/references/{project_id}/{reference_id}` CRUD for a reference (also, PATCH and DELETE)

This convention will also be used for the chat:

 - `POST api/ai/chat/{project_id}`

cguedes (approved):

PR #396, opened 2023-08-22
feat: Create python http equivalent of settingsManager.ts

### LINKED ISSUE #364: Create Python/HTTP equivalent of settingsManager.ts
Part of work for #347

gjreda:
fixes #364
_____

I think this generic API for settings should be all we need, but let me know otherwise @cguedes @danvk


python/sidecar/http.py
```diff
@@ -218,3 +221,17 @@ async def delete_file(project_id: str, filepath: Path):
...
+
+# Settings API
+# --------------
+@settings_api.get("/")
+async def get_settings() -> SettingsResponse:
+    user_id = "user1"
+    return settings.get_settings_for_user(user_id)
+
+
+@settings_api.post("/")
```
cguedes, line 234: This should be PUT instead of POST as this will update settings instead of creating settings.

python/sidecar/typing.py
```diff
@@ -201,6 +201,14 @@ class SearchResponse(RefStudioModel):
...
 
 
+class UpdateSettingsRequest(RefStudioModel):
+    settings: dict[str, Any]
+
+
+class SettingsResponse(RefStudioModel):
+    settings: dict[str, Any]
+
+
```
cguedes, line 211: We know the structure of the settings. Can't we have a strongly typed definition?

cguedes (changes requested):

python/sidecar/settings.py
```diff
@@ -38,4 +41,53 @@
 handler.setFormatter(formatter)
 
 logger.addHandler(handler)
-logger.disabled = os.environ.get("SIDECAR_ENABLE_LOGGING", "false").lower() == "true"
\ No newline at end of file
+logger.disabled = os.environ.get("SIDECAR_ENABLE_LOGGING", "false").lower() == "true"
+
+
+def make_settings_json_path(user_id: str) -> Path:
+    filepath = Path(WEB_STORAGE_URL / "settings.json")
```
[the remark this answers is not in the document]
gjreda: Yep, good catch

cguedes (approved):

PR #407, opened 2023-08-23
Sidecar server

### LINKED ISSUE #406: Use the Python HTTP API from Tauri
We currently communicate from TypeScript <-> Python using a [sidecar process](https://tauri.app/v1/guides/building/sidecar/). We start a new process whenever we need to communicate. data is sent to Python as a JSON command-line argument and we read JSON back from stdout.

With the web version of RefStudio (#347) we have an HTTP server. This means that the web version and the desktop version have two different ways of communicating. Ideally we'd make the desktop app use the HTTP server as well so that we can unify these code paths.

See @hammer's comment here for some context: https://github.com/refstudio/refstudio/issues/17#issuecomment-1563288836

The simplest path forward is to use `Command` to spawn the web server and then Tauri's version of [fetch](https://tauri.app/v1/api/js/http#fetch) to communicate with it.

danvk:
See #406 

This PR adds a `serve` subcommand to the Python sidecar that runs a web server. Eventually this should be the only command, but for now it coexists with existing sidecar commands. This will let us share much more code between the web and desktop versions of RefStudio.

When the app starts, it uses `Command.spawn` to run the `serve` command, starting a web server. This server's port is in the allowlist in `tauri.conf.json` which allows us to connect to it using Tauri's version of `fetch`.

A few notes:

- The server's stderr/stdout are forwarded to the dev console so you can see them, but we could consider putting them in a file (either using the Tauri FS API or something in Pythonland) if this is too noisy.
- When you quit the app, Tauri closes the server process.
- When you run in debug mode and make changes to the TypeScript code, Tauri does not kill the server. This means that when we try to start a new server from TypeScript, it will fail due to a port collision. We may want to think more about this, but for now, I'm detecting this situation and killing the old server via a `/api/meta/shutdown` endpoint.
- uvicorn is pretty good about keeping the server running when Python throws an exception. But if the process dies, we'll restart it from TypeScript.

Forwarded logs from the server:

image

Killing the server triggers a restart, after which the API is still usable:

https://github.com/refstudio/refstudio/assets/98301/fb7b0a90-80af-4b5b-817e-3adf26f30a58


src/api/sidecar.ts
```diff
@@ -81,3 +95,106 @@ export async function callSidecar<T extends keyof CliCommands>(
...
+
+(window as any).startServer = startServer;
+(window as any).killServer = killServer;
+(window as any).isServerHealthy = isServerHealthy;
+(window as any).getProjects = async () => {
+  const response = await tauriFetch('http://localhost:1487/api/projects');
+  console.log(response.data);
+};
+
+export async function getServer() {
```
cguedes, line 166: Move this server startup code to a new file.

src/api/server.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/**
+ * Ensure there is exactly one RefStudio server running.
+ *
+ * Repeated calls have no effect. The server will terminate when the RefStudio app terminates.
+ * If an existing server not managed by us is already running, it will be killed. (This can happen
+ * when you run Tauri Desktop in dev mode and make TypeScript changes.)
+ */
+export function runRefStudioServerOnDesktop() {
```
danvk, line 104: There's a possible race condition where we start the server and try to fetch something from it (maybe settings) before it's ready to respond. I'm hoping this won't come up in practice. If it is, we'll want to expose something like `function waitForServer(): Promise<void>` and `await` that before running `tauriFetch`. But this will introduce another difference between desktop and web (where you can assume the server is always running) so I'd really like to avoid that if we can.

src/api/server.ts
```diff
@@ -0,0 +1,114 @@
...
+  state: 'starting';
+}
+interface ReadyState {
+  state: 'ready';
+  process: Child;
+}
+type ServerState = StartingState | ReadyState;
+
+// A null value means that no one has requested a server yet.
+let serverProcess: ServerState | null = null;
```
danvk, line 18: Lots of ceremony here but I really want to make sure that multiple calls to `getOrStartServer` don't result in multiple servers being started.

danvk (commented):
Let me know your thoughts about testing. It's important code, but the test is going to involve enough mocks that I'm not sure it will really be that helpful.

src/api/server.ts
cguedes: I think it's ok. AppStartup (and the splashscreen) is there to ensure we don't call any other API (or render the app) before everything is initialised.

python/sidecar/typing.py
```diff
@@ -198,7 +198,11 @@ class S2SearchResult(RefStudioModel):
 class SearchResponse(RefStudioModel):
     status: ResponseStatus
     message: str
-    results: list[S2SearchResult] 
+    results: list[S2SearchResult]
+
+
+class ServeResponse(RefStudioModel):
+    pass
```
cguedes, line 205: What's this @danvk? I don't see this `ServeResponse` type being used in the API.

src/api/server.ts
cguedes: It's ok.

cguedes (approved):
Just left a comment about a typing definition. 

I agree that making (unit) tests for this would be more complex than the value we get. This is something we can test quickly as we build e2e tests for the app!

src/api/server.ts
danvk: Yeah, I think we're going to need some kind of waiting logic. Here's what I see in the debug logs. Note that we make at least 3 API calls (init settings, ingest references, read file tree) before the HTTP server reports that it's listening:
image

cguedes (approved):

PR #417, opened 2023-08-24
Universal settings

danvk:
Part of the work for #347 

This PR includes
 - Desktop app wait for the server to be ready before start (during splash-screen, in AppStartup)
 - Enhanced codegen for API types (from python)
 - Removed (funcional) dependency on `tauri-settings`
 - Adopt the new settings 
 - Universal GET and POST functions to make HTTP requests via tauri (desktop) or native fetch (web)

Pending:
 - Review structure (schema) of settings and try to make it a key/val store instead of the current object tree
 - Remove dependency of `tauri-settings`
 - (for Dev) Automatically load .env settings (as defaults) on server startup

image

Edit: Description updated by @cguedes


src/AppStartup.tsx
```diff
@@ -35,7 +40,7 @@ export function AppStartup() {
 
         if (isMounted()) {
           setInitialized(true);
-          const projectDir = getCachedSetting('project.currentDir');
+          const projectDir = getCachedSetting('project.current_directory');
```
danvk, line 43: Python & TS share types for `SettingsSchema` now so there are a few changes like this. If we really want to camelcase everything then we should do it with some kind of middleware. But I'd rather have fewer types.

src/api/server.ts
```diff
@@ -2,39 +2,52 @@
...
 // A null value means that no one has requested a server yet.
 let serverProcess: ServerState | null = null;
 
-async function startServer() {
+async function startServer(servingCallback: () => void) {
   console.log('Starting RefStudio server');
   const command = new TauriCommand('call-sidecar', ['serve']);
   command.stderr.addListener('data', (line) => {
     console.log('server stderr', line);
+    if (line.includes('running on http')) {
```
danvk, line 34: This matches:

    Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)

Definitely open to other ideas on how to tell that the app is listening for connections!

src/settings/settingsManager.ts
```diff
@@ -1,40 +1,18 @@
...
-  sidecar: {
-    logging: {
-      active: boolean;
-      path: string;
-    };
-  };
-}
+type DeepRequired<T> = T extends object ? { [k in keyof T]-?: DeepRequired<T[k]> } : T;
+
+export type SettingsSchema = DeepRequired<apiTypes.SettingsSchema>;
```
danvk, line 15: Chatted about this briefly with @gjreda. This is going to be an issue with any Pydantic classes with default values, i.e. most of our current API types. These get modeled on the TS side as optional properties. For the settings PUT request body, I think this may be what we want (you don't have to set all the settings at once) but for the response type it isn't. All fields are required on the `/api/settings/` response. I've papered over that here but ultimately we'll want to have distinct request/response types coming from the Python side.

src/settings/settingsManager.ts
```diff
@@ -48,67 +26,31 @@ export type SettingsManagerView = Pick<
...
+
+  settingsManager = {
+    default: settings,
+    getCache: (key) => getDotNotation(settings, key)!,
+    setCache: (key, value) => {
+      setDotNotation(settings, key, value);
+      return value;
+    },
+    syncCache: async () => {
+      const newSettings = (await universalPost('/api/settings/', settings, 'PUT')) as SettingsSchema;
```
danvk, line 49: Note: the trailing slash is important. If you leave it off, you get confusing CORS errors from the dev server in web mode.

src/settings/settingsManager.ts
danvk, line 1: This file gets a lot simpler and we get to drop the (runtime) dependency on tauri-settings, which is a big win. We could drop the dependency on its types in a follow-up.

src/settings/__tests__/settings.test.ts
danvk, line 1: I deleted these this to get tests passing, but we should update/rewrite them instead.

src/api/api.ts
danvk, line 1: This papers over the differences between Tauri fetch and web fetch.

src/api/api-types.ts
```diff
@@ -0,0 +1,324 @@
...
+ * via the `definition` "ProjectSettings".
+ */
+export interface ProjectSettings {
+  current_directory?: string;
+}
+/**
+ * This interface was referenced by `ApiSchema`'s JSON-Schema
+ * via the `definition` "SettingsSchema".
+ */
+export interface SettingsSchema {
```
danvk, line 313: These are all the types that appear in a FastAPI request/response type or are referenced from one of those types. Note how all the fields are optional, which is not what we want.

danvk (commented):
This is as far as I got today. It includes logic to wait for the server to start up (for the desktop app), use the settings API (for both desktop and web) and the start of codegen for our HTTP API.

python/sidecar/typing.py
```diff
@@ -202,7 +207,7 @@ class SearchResponse(RefStudioModel):
 
 
 class OpenAISettings(RefStudioModel):
-    api_key: str = ""
+    api_key: str = os.environ.get("OPENAI_API_KEY")
```
danvk, line 210: This is where the key leaks -- when this schema gets converted to JSONSchema, the default value is serialized, leaking the key. TS doesn't have any notion of default values, so it goes away in `api-types.ts`.

src/api/api.ts
cguedes: Thanks for starting this one.

src/api/api.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+    return response.data as ResponsePayload;
+  }
+}
+
+/** Issue a POST request with a JSON payload using either web fetch or Tauri fetch */
+export async function universalPost<ResponsePayload = unknown, RequestPayload = unknown>(
+  path: string,
+  payload: RequestPayload,
+  method: HttpVerb = 'POST',
```
sehyod, line 26: nit: we should not be able to pass a method, or we should rename the function to `universalFetch`

sehyod (approved):

src/api/api.ts
cguedes: Ok. Will do that. Agree.

cguedes (approved):

PR #419, opened 2023-08-24
chore: Add black and format python code

gjreda:
fixes https://github.com/refstudio/refstudio/issues/416

Except for changes to the workspace .vscode/settings.json and .husky, this PR is entirely formatting changes.

PR #425, opened 2023-08-28
Universal desktop web codebase

### LINKED ISSUE #422: Universal Desktop + Web codebase
Related to #347 

Now that #393 is implemented, we should unify the codebase between desktop and web and clean-up the code to take advantage of the [sidecar server](https://github.com/refstudio/refstudio/pull/407).

cguedes:
This PR closes #422 by removing the if/else conditions using `import.meta.env.VITE_IS_WEB` to implement a different code path between the desktop and web applications.

Changes:
- ✅ Updated backend API to return project name
- ✅ Updated backend API to ensure required files and directories (`.staging/references.json` and the project folder)
- ✅ Updated new/open project flow to always use the backend API
- ✅ Updated backend API to support HEAD `/api/fs/... ` (for file exists)
- ✅ Update frontend to support the sample project
     - ℹ️ The app ensure we only have a single project for the sample (checking the project name), and also overrides sample text files, but don't override the project files (ex: uploaded references)
- ✅ Save the project ID in settings instead of the project path
     - ℹ️ 
- ✅ Update filesystem to always use the backend API instead of tauri/fs methods
- ✅ Fix upload & save files for the desktop Tauri (~file upload don't work~) 🎉 
- ✅ Adopt backend API instead of sidecar
    - ✅ chat
    - ✅ ingestion
    - ✅ rewrite
    - ✅ completion

Follow-up:
 - #445 
 - #446
 - #447
 - #448
 - #449
 - #454
 - #455 
 - #456


sehyod (changes requested):

src/api/projectsAPI.ts
```diff
@@ -1,41 +1,145 @@
...
+type ProjectPostResponse = ProjectsResponse;
 
-export async function getRemoteProjects() {
-  const res = await fetch(`/api/projects/`);
-  const data = res.json() as unknown as ProjectsResponse;
-  return data;
+// ########################################################################################
+// PROJECT LIFE CYCLE
+// ########################################################################################
+export async function readAllProjects() {
```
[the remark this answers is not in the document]
cguedes: Sure!

src/api/projectsAPI.ts
```diff
@@ -1,41 +1,145 @@
...
+  return Object.keys(projects).map(
+    (id) => ({ id, path: projects[id].project_path, name: projects[id].project_name } as ProjectInfo),
+  );
 }
-
-export async function getRemoteProject(projectId: string) {
-  const res = await fetch(`/api/projects/${projectId}`);
-  const data = res.json() as unknown as ProjectIdResponse;
-  return data;
+export async function readProjectById(projectId: string) {
```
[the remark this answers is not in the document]
cguedes: Sure!

src/api/projectsAPI.ts
```diff
@@ -1,41 +1,145 @@
...
-    method: 'POST',
-    // We need to update the API definition in order to use BODY
-    // https://stackoverflow.com/questions/59929028/python-fastapi-error-422-with-post-request-when-sending-json-data
-    // body: JSON.stringify({ project_name: projectName }),
-    headers: {
-      Accept: 'application/json',
-    },
-  });
-  const payload = (await response.json()) as ProjectsResponse;
+export async function createRemoteProject(projectName: string, projectPath?: string) {
```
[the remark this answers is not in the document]
cguedes: Ok

python/generate_schema.py
```diff
@@ -2,9 +2,8 @@
 import json
 
 from fastapi.openapi.utils import get_openapi
-from starlette.routing import Mount
-
 from sidecar.typing import CliCommands
+from starlette.routing import Mount
```
danvk, line 6: This is more a question for @gjreda and #26 but I've had good luck using [isort](https://pycqa.github.io/isort/) to enforce a sort order for imports in Python in the past. The usual grouping is: standard library imports, third-party imports, your own imports.

python/sidecar/http.py
```diff
@@ -81,17 +80,19 @@ async def http_search_s2(req: SearchRequest) -> SearchResponse:
 @ai_api.post("/rewrite")
 async def http_ai_rewrite(
     req: RewriteRequest,
-    user_settings: Annotated[SettingsSchema, Depends(settings.get_settings_for_user)],
+    # user_settings: Annotated[SettingsSchema, Depends(settings.get_settings_for_user)],
```
danvk, line 83: delete these lines if they're no longer needed

python/sidecar/projects.py
```diff
@@ -57,7 +56,7 @@ def update_project_path_storage(
             "project_name": project_name,
             "project_path": str(project_path),
         }
-        json.dump(data, f)
+        json.dump(data, f, indent=2, default=str)
```
danvk, line 59: `indent=2` is fine but `default=str` makes me nervous. Do we need this?

python/sidecar/projects.py
```diff
@@ -67,11 +66,17 @@ def get_project_path(user_id: str, project_id: str) -> Path:
     return Path(data[project_id]["project_path"])
 
 
+def get_project_name(user_id: str, project_id: str) -> Path:
```
danvk, line 69: Why is the name a `Path`? I'd expect a `str`.

src/api/api.ts
```diff
@@ -1,69 +1,175 @@
...
+  if (!response.ok) {
+    throw new Error(`Failed to ${method} ${path}: ${response.status} - ${JSON.stringify(response.data)}`);
+  }
+  switch (responseParser) {
+    case 'JSON':
+      console.log('TAURI FETCH RESPONSE DATA', response.data);
+      return response.data as ResponsePayload;
+    case 'ArrayBuffer':
+      console.log('TAURI FETCH RESPONSE DATA', response.data);
+      return response.data as ResponsePayload;
```
danvk, line 142: So `response.data` is already an `ArrayBuffer` or a parsed JSON object depending on the response type? Interesting API!

src/api/api.ts
```diff
@@ -1,69 +1,175 @@
...
-    }
-    return response.json() as ResponsePayload;
-  } else {
-    const response = await tauriFetch(REFSTUDIO_HOST + path);
-    if (!response.ok) {
-      throw new Error(`Failed to get ${path}: ${response.status}`);
-    }
-    return response.data as ResponsePayload;
-  }
+export async function universalGet<ResponsePayload = unknown>(
```
danvk, line 9: Two things here:

1. `unknown` is the "default default" for type parameters in TS, so there's no need to write "= unknown" explicitly.
2. This is an example of a so-called [return-only generic](https://effectivetypescript.com/2020/08/12/generics-golden-rule/). These are generally frowned upon in TS because they don't actually give you any type safety -- you can claim the return type is whatever you want. The better way would be to have this always return `Promise<unknown>` and then do `await universalGet('...') as ResponsePayload` in the calling code since at least that way it's explicit that you're doing a type assertion.

That being said, I think this will all change when we plug in code to generate these types so fine to leave as-is for now.

src/api/api.ts
```diff
@@ -1,69 +1,175 @@
...
+/** Issue a PATCH request with a JSON payload using either web fetch or Tauri fetch */
+export async function universalPatch<ResponsePayload = unknown, RequestPayload = unknown>(
+  path: string,
+  payload: RequestPayload,
+  responseParser: ResponseParser = 'JSON',
+): Promise<ResponsePayload> {
+  return universalRequest('PATCH', path, payload, responseParser);
+}
+
+/** Issue a HEAD request using either web fetch or Tauri fetch */
```
danvk, line 52: Document that the returned value is the response code.

src/api/api.ts
```diff
@@ -1,69 +1,175 @@
...
+    return fetch(path, { method: 'PUT', body: formData });
+  }
+
+  // TAURI
+  const url = REFSTUDIO_HOST + path;
+  const buffer = typeof content === 'string' ? new TextEncoder().encode(content) : new Uint8Array(content);
+  const body = Body.form({
+    file: {
+      // either a path or an array buffer of the file contents.
+      // We can't use a path bc Tauri check if file exists and has read permissions. Also, will send the file content.
```
danvk, line 160: Is there a bug you can reference here? This is surprising behavior.

src/api/projectsAPI.ts
```diff
@@ -1,41 +1,143 @@
...
+      name: 'uploads',
+      children: uploadsFiles,
+    },
+  ];
+}
+
+// ########################################################################################
+// PROJECT FILE SYSTEM
+// ########################################################################################
+export const writeProjectTextFile = async (projectId: string, relativeFilePath: string, contents: string) => {
```
danvk, line 91: Can simplify to:

```ts
export const writeProjectTextFile = async (projectId: string, relativeFilePath: string, contents: string) =>
  universalPutFile(`/api/fs/${projectId}/${relativeFilePath}`, relativeFilePath, contents);
```

and also for the next method, unless it's intentional that you want to swallow the return value of `universalPutFile`.

src/api/projectsAPI.ts
```diff
@@ -1,41 +1,143 @@
...
+  const status = await universalHead(`/api/fs/${projectId}/${filePath}`);
+  return status === 200;
+};
+
+export const deleteProjectFile = async (projectId: string, relativeFilePath: string) => {
+  await universalDelete(`/api/fs/${projectId}/${relativeFilePath}`);
+};
+
+export const renameProjectFile = async (projectId: string, relativeFilePath: string, newFilePath: string) => {
+  const fileExists = await existsProjectFile(projectId, relativeFilePath);
```
danvk, line 130: Do we have an issue to implement a real rename endpoint?

src/api/sidecar.ts
```diff
@@ -1,64 +1,10 @@
...
 import { CliCommands } from './types';
 
-interface SharedCommand {
-  execute: typeof TauriCommand.prototype.execute;
-}
-interface StubCommandOptions {
-  env: Record<string, string>;
-}
-
-class StubCommand implements SharedCommand {
```
danvk, line 15: 🎉

src/application/listeners/projectEventListeners.ts
```diff
@@ -1,81 +1,99 @@
...
 
 export function useFileProjectNewSampleListener() {
   const sampleProject = useSetAtom(newSampleProjectAtom);
+  const createFile = useSetAtom(createFileAtom);
   return async () => {
-    const newProjectPath = `${await getNewProjectsBaseDir()}RefStudio Sample`;
-    await sampleProject(newProjectPath);
-    persistProjectDirInSettings(newProjectPath);
-    notifyInfo('New project created at ' + newProjectPath);
+    const projects = await readAllProjects();
```
danvk, line 49: should this logic (create or return sample project) live in the backend?

src/application/listeners/projectEventListeners.ts
```diff
@@ -1,81 +1,99 @@
...
-    }
-
-    const selectedPath = await open({
-      directory: true,
-      multiple: false,
-      defaultPath: await getNewProjectsBaseDir(),
-      title: 'Open RefStudio project',
-    });
+      const projects = await readAllProjects();
+      projectId = projects[0].id;
```
danvk, line 70: is there a guarantee that there will be a project?

src/application/listeners/projectEventListeners.ts
```diff
@@ -1,81 +1,99 @@
...
+        defaultPath: await getNewProjectsBaseDir(),
+        title: 'Open RefStudio project',
+      });
 
-    if (typeof selectedPath === 'string') {
-      await openProject(selectedPath);
-      persistProjectDirInSettings(selectedPath);
-      notifyInfo('New project open at ' + selectedPath);
+      if (typeof selectedPath === 'string') {
+        const projects = await readAllProjects();
```
danvk, line 80: I don't feel too strongly either way, but this logic could also go in the backend (like a projectIdForPath endpoint).

src/application/listeners/projectEventListeners.ts
```diff
@@ -1,81 +1,99 @@
...
+        projectId = projectWithSamePath.id;
+      } else {
+        notifyInfo('User canceled operation to open new project');
+        return;
+      }
     }
+
+    const projectToOpen = await readProjectById(projectId);
+    await openProject(projectId, projectToOpen.path, projectToOpen.name);
+    persistProjectIdInSettings(projectId);
```
danvk, line 95: Could we name this something like `persistCurrentProject` or `persistActiveProject`? Is there a way we could avoid having to call it in so many places? Every time you open a project, we should persist it.

src/io/filesystem.ts
```diff
@@ -132,128 +61,54 @@ export function makeUploadPath(filename: string) {
...
 }
 
-export function isInUploadsDir(relativePath: string) {
-  return relativePath.startsWith(`${sep}${UPLOADS_DIR}`);
+export function isInUploadsDir(relativeFilePath: string) {
+  return relativeFilePath.startsWith(`${UPLOADS_DIR}${sep}`);
 }
 
 export async function uploadFiles(systemFiles: File[]) {
-  if (import.meta.env.VITE_IS_WEB) {
```
danvk, line 140: Nice to see all these `if` statements go away!

src/io/filesystem.ts
```diff
@@ -279,62 +134,45 @@ async function convertTauriFileEntryToFileEntry(entry: TauriFileEntry): Promise<
...
 // #####################################################################################
 // file path utils
+//
+// NOTE: We decided that ref studio paths should always start WITHOUT a leading /.
+//       file.txt -> file.txt
 // #####################################################################################
-// file.txt -> /file.txt
-export function makeRefStudioPath(fileName: string) {
-  return `${sep}${fileName}`;
+export function makeRefStudioPath(filePath: string) {
```
danvk, line 141: can we just get rid of this function?

src/io/filesystem.ts
```diff
@@ -344,91 +182,76 @@ export async function readFileContent(file: FileFileEntry): Promise<EditorConten
...
-      const newSystemPath = await join(getParentFolder(systemPath), newName);
-      if (await exists(newSystemPath)) {
-        console.warn(`Another file with the same name already exists: ${newSystemPath}`);
-        return { success: false };
-      }
-      await tauriRenameFile(systemPath, newSystemPath);
-      const newRelativePath = getRefStudioPath(newSystemPath);
-      return { success: true, newPath: newRelativePath };
+    if (await existsProjectFile(currentProjectId, newRelativePath)) {
+      console.warn(`Another file with the same name already exists: ${newRelativePath}`);
```
danvk, line 208: shouldn't this check go in `renameProjectFile`?

danvk (dismissed):
Lots of comments but overall looks great!

python/sidecar/projects.py
```diff
@@ -67,11 +66,17 @@ def get_project_path(user_id: str, project_id: str) -> Path:
...
+    data = read_project_path_storage(user_id)
+    return Path(data[project_id]["project_name"])
```
gjreda, line 71:
```suggestion
def get_project_name(user_id: str, project_id: str) -> str:
    data = read_project_path_storage(user_id)
    return data[project_id]["project_name"]
```

python/sidecar/projects.py
gjreda, line 59:
```suggestion
        json.dump(data, f, indent=2)
```

python/generate_schema.py
gjreda: @danvk I'm a fan of that pattern and thought I'd set it up as part of #419, but maybe I missed a piece.

python/sidecar/http.py
cguedes: Right. I had them to discuss with Greg.

.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
...
+    "version": "0.2.0",
+    "configurations": [
+      {
+        "name": "Python: FastAPI",
+        "type": "python",
+        "request": "launch",
+        "module": "uvicorn",
+        "cwd": "${workspaceFolder}/python",
+        "args": ["web:api", "--reload"],
+        "jinja": true,
```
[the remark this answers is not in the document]
cguedes: This was added by default when I (and Greg) created this configuration file. I see this is related with Flask. So will remove.

python/sidecar/http.py
```diff
@@ -298,6 +302,16 @@ async def read_file(project_id: str, filepath: Path):
...
 
 
+@filesystem_api.head("/{project_id}/{filepath:path}", status_code=200)
+async def head_file(project_id: str, filepath: Path):
+    user_id = "user1"
+    project_path = projects.get_project_path(user_id, project_id)
+    filepath = project_path / filepath
+
+    if not filepath.exists():
+        raise HTTPException(status_code=404, detail="File not found")
```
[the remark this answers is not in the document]
cguedes: HEAD requests don't have a response body.

src/io/filesystem.ts
cguedes: that's correct

src/api/projectsAPI.ts
cguedes: We discussed having that endpoint. But it won't fit into the filesystem API we designed. Maybe we can have a specific enpoint in the project API in the future, but not for now IMO.

src/application/listeners/projectEventListeners.ts
cguedes: Nice catch! Will throw an error in no project exists.

src/application/listeners/projectEventListeners.ts
cguedes: We can call it diferently but I think we can't avoid calling in these 3 places (new, open, sample).

src/application/listeners/projectEventListeners.ts
cguedes: I see your point, but I would rather implement a filter by path in the existing API /projects. For now I think it's fine how it is.

src/io/__tests__/filesystem.test.ts
```diff
@@ -325,47 +247,49 @@ describe('filesystem', () => {
...
-      vi.mocked(tauriExists).mockResolvedValue(false);
-      const result = await renameFile('/some/relative.txt', 'updated.txt');
-      expect(result).toStrictEqual({ success: true, newPath: '/some/updated.txt' });
-      expect(tauriExists).toHaveBeenCalledTimes(1);
-      expect(tauriRenameFile).toHaveBeenCalledTimes(1);
-      expect(tauriRenameFile).toHaveBeenCalledWith(
-        await getSystemPath('/some/relative.txt'),
-        await getSystemPath('/some/updated.txt'), //
-      );
+    it.skip('should rename file content via projects API', async () => {
```
danvk, line 270: Do you want to commit these as skipped? Should we delete them?

danvk (approved):

gjreda (dismissed):

src/io/__tests__/filesystem.test.ts
cguedes: I will fix the rename regression and the tests in a follow-up PR.

src/api/api.ts
cguedes: Ok, thanks for sharing this information. I leave as-is for now bc we will update this soon.

src/api/api.ts
```diff
@@ -1,69 +1,175 @@
...
 
 /** Issue a POST request with a JSON payload using either web fetch or Tauri fetch */
-// Note: Comment this out for the first need of POST. We need to comment this out because of Knip.
-// export async function universalPost<ResponsePayload = unknown, RequestPayload = unknown>(
-//   path: string,
-//   payload: RequestPayload,
-// ): Promise<ResponsePayload> {
-//   return universalRequest('POST', path, payload);
-// }
+export async function universalPost<ResponsePayload = unknown, RequestPayload = unknown>(
```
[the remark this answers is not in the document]
cguedes: Sure. That's a solution for this.

src/api/api.ts
```diff
@@ -1,69 +1,175 @@
...
+      : {
+          Accept: 'application/json',
+          'Content-Type': 'application/json',
+        };
+  const response = await fetch(path, {
+    method,
+    headers,
+    body: payload instanceof FormData ? payload : payload ? JSON.stringify(payload) : undefined,
+  });
+  console.log('WEB FETCH RESPONSE STATUS:', response.status);
```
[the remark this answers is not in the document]
cguedes: I wasn't sure. Maybe now we can do that 😅

src/api/api.ts
cguedes: I know!!

src/api/api.ts
cguedes: I don't have it. I've got 422 Unprocessable Entity from the python code when this field was missing.
I will reproduce tomorrow and send you.

src/application/listeners/projectEventListeners.ts
cguedes: Most likely yes, with a speacialized API for sample projects.

src/api/projectsAPI.ts
cguedes: Ok. Fixed this.

src/api/projectsAPI.ts
```diff
@@ -1,41 +1,143 @@
...
+    id: projectId,
+    path: project_path,
+    name: project_name,
   };
 }
+
+// ########################################################################################
+// PROJECT FILE STRUCTURE
+// ########################################################################################
+/** NOTE: This is a simplified version of the function */
```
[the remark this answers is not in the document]
cguedes: That only works for one deep level of files. We will fix this in #448

python/sidecar/projects.py
cguedes: Bc I just copy/pasted python code! This was fixed already.

cguedes (commented):
Replied to all comments.

danvk (approved):
Very excited to get this in!

PR #432, opened 2023-08-28
Add api for search

### LINKED ISSUE #428: Backend adds API for `/search`

gjreda:
fixes #428

### DIFF
--- a/python/web.py
+++ b/python/web.py
@@ -1,17 +1,16 @@
 import uvicorn
 from fastapi import FastAPI
-
 from sidecar import http
 
-
 api = FastAPI()
-api.mount("/api/sidecar", http.sidecar_api)
-api.mount("/api/references", http.references_api)
-api.mount("/api/ai", http.ai_api)
-api.mount("/api/fs", http.filesystem_api)
-api.mount("/api/projects", http.project_api)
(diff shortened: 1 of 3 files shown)


python/web.py
```diff
@@ -1,17 +1,16 @@
...
-
 api = FastAPI()
-api.mount("/api/sidecar", http.sidecar_api)
-api.mount("/api/references", http.references_api)
-api.mount("/api/ai", http.ai_api)
-api.mount("/api/fs", http.filesystem_api)
-api.mount("/api/projects", http.project_api)
-api.mount("/api/meta", http.meta_api)
-api.mount("/api/settings", http.settings_api)
+api.mount("/api/sidecar", http.sidecar_api, name="sidecar")
```
cguedes, line 6: @gjreda why do we have this name parameter?

cguedes (approved):
All good. Just a comment about the name parameter.

python/web.py
gjreda: Starlette optionally uses it for some stuff under the hood. I thought it might show up in the logs, but alas it doesn't.

Anyway, I just decided to leave it for now as I suspect it might be handy later on.

Issue #415, opened 2023-08-24
Break up http.py


gjreda:
This has grown quite large and should be broken up into submodules

gjreda:
@cguedes I'm planning to restructure the backend and roughly follow [this template](https://github.com/zhanymkanov/fastapi-best-practices), grouping things by domain. So we'll have something like the following:
```
sidecar/
    config.py --> config-related stuff from what is now called settings.py
    main.py
    web.py
    search/
        router.py --> definition of endpoints, stuff that is now in http.py
        schemas.py --> definition of search-related types, from typing.py
        service.py --> business logic sort of code, stuff that now lives in search.py
    projects/
        router.py --> endpoints from http.py
        schemas.py
        services.py
    references/ --> i'll probably rename `ingest` related logic to `references`, since that's the endpoint prefix we use
        ....
....
```
This pattern largely mirrors what Netflix does for [Dispatch](https://github.com/Netflix/dispatch/tree/master). I think it'll be nice to have as other domains of the application get built out, and also as existing areas like `ai` get more depth to them.

Let me know what you think.

gjreda:
Gives us some nice organization to the swagger docs too
<img width="1224" alt="image" src="https://github.com/refstudio/refstudio/assets/921995/099ea0f5-292a-43ec-b7fc-8bbc9a36a001">

cguedes:
I like this a lot. 💪

PR #457, opened 2023-08-30
Generate a TypeScript API from our OpenAPI spec

danvk:
This PR sets up [opeanapi-typescript](https://openapi-ts.pages.dev/introduction/) to generate types from our OpenAPI spec. I also looked at [openapi-typescript-codegen](https://github.com/ferdikoomen/openapi-typescript-codegen) but it generates an enormous number of files and classes for each API, which isn't a good fit for how we work.

There are a few complications:

1. We actually have several APIs (settings, fs, projects, etc.). Having separate TS files for each of these would be a pain. To get a single TypeScript file, I had to merge all our OpenAPI specs.
2. The resulting API schema types are all defined in one big, nested interface. You can see what that looks like [here](https://github.com/refstudio/refstudio/pull/457/files#diff-3ad856bdae050e287f29155e86b9bc8d7ad818d790f3af6b19b973bce16322fcR162). This isn't quite as nice as having all these types as distinct, top-level `interface`s like we do now. I toyed with [forwarding](https://github.com/refstudio/refstudio/pull/457/commits/1025cf7a3a84e2fba7b2e731da903899133e3067#diff-d75a130d0576d6848eb8a701b4fdde5777c3fada17eff2caed62638fd2c48355R4) all the types, but this doesn't look as nice in your editor and you lose things like JSDoc. So I decided to keep our existing json2ts types (`api-types.ts`). Which leads us to…
3. I wanted to patch `api-paths.ts` to use the types from `api-types.ts`. I thought this would be find and replace, but they generate type names in ever so slightly different ways. So I wound up setting up a Python script to do some normalization and patch `api-paths.ts`.

So now we have types for requests / response and the various paths in our API available in TypeScript.

There's also [openapi-fetch](https://openapi-ts.pages.dev/openapi-fetch/), which generates a runtime API for you with nice types. You can even pass your own `fetch` function to it. Unfortunately, it's [typed as `typeof fetch`](https://github.com/drwpow/openapi-typescript/blob/1ee5d2e43a4b4dbb8160afacfa17fc1d80714745/packages/openapi-fetch/src/index.ts#L18), which means that we'd need to make a web `fetch`-compatible version of Tauri's fetch. Yuck. I tried going down this path a bit but didn't like where I wound up.

Instead, I dropped openapi-fetch and decided to add types to the existing `universalGet` / `universalPost` / etc. APIs. This does mean some reinventing the wheel for what openapi-fetch does, but it's nice that we get to reuse the code that @cguedes already set up in #425.

There are still quite a few type assertions here because our API (in Python) is missing quite a few types.

In any event, you get nice types!
image

And autocomplete! (These are just the paths with POST endpoints.)
image

Still to do:

- [x] Write some unit tests for Typed API
- [ ] Add the other HTTP methods (PUT, PATCH, DELETE)
- [ ] Update remaining API calls


cguedes:
@danvk everything looks nice to me. That TS code on `src/api/typed-api.ts` is 🔥

cguedes:
@danvk with the combined swagger spec we can now use this code to access the full api documentation via [http://127.0.0.1:8000/docs](http://127.0.0.1:8000/docs).

```python
api = FastAPI(swagger_ui_parameters={"url": "/combined.openapi.json"})


@api.get("/combined.openapi.json")
def read_openapi_json_file():
    return FileResponse("combined.openapi.json")
```

image

CC @gjreda

gjreda:
@cguedes I don't think that will be necessary once I complete #415. See more here: https://fastapi.tiangolo.com/tutorial/bigger-applications/#check-the-automatic-api-docs

cguedes:
> @cguedes I don't think that will be necessary once I complete https://github.com/refstudio/refstudio/issues/415. See more here: https://fastapi.tiangolo.com/tutorial/bigger-applications/#check-the-automatic-api-docs

So I think @danvk won't need to merge the open API specs into a single file.

danvk:
That does seem like a nicer way to get a combined OpenAPI but I think I'll leave that for another PR. We can remove my merging code once we do that.

src/api/chat.ts
```diff
@@ -1,12 +1,12 @@
...
-import { ChatRequest, ChatResponse } from './api-types';
+import { apiPost } from './typed-api';
 
 export async function chatWithAI(projectId: string, text: string): Promise<string[]> {
   try {
     if (!text) {
       return [];
     }
-    const res = await universalPost<ChatResponse, ChatRequest>(`/api/ai/${projectId}/chat`, { text, n_choices: 3 });
+    // const res = await universalPost<ChatResponse, ChatRequest>(`/api/ai/${projectId}/chat`, { text, n_choices: 3 });
```
cguedes, line 8: do you want to keep this comment?

src/api/completion.ts
```diff
@@ -1,14 +1,14 @@
...
+import { apiPost } from './typed-api';
 
 export async function completeSentence(text: string): Promise<string[]> {
   try {
     if (text.trim() === '') {
       return [];
     }
 
-    const response = await universalPost<TextCompletionResponse, TextCompletionRequest>('/api/ai/completion', {
+    // const response = await universalPost<TextCompletionResponse, TextCompletionRequest>('/api/ai/completion', {
```
cguedes, line 10: same, consider removing this.

src/api/projectsAPI.ts
```diff
@@ -22,11 +23,15 @@ type ProjectPostResponse = ProjectsResponse;
...
 // ########################################################################################
 export async function readAllProjects(): Promise<ProjectInfo[]> {
-  const projects = await universalGet<ProjectsResponse>(`/api/projects/`);
+  const projects = (await apiGetJson('/api/projects/')) as ProjectsResponse;
+  // const projects = await universalGet<ProjectsResponse>(`/api/projects/`);
   return Object.keys(projects).map((id) => ({ id, path: projects[id].project_path, name: projects[id].project_name }));
 }
 export async function readProjectById(projectId: string): Promise<ProjectInfo> {
-  const projectInfo = await universalGet<ProjectGetResponse>(`/api/projects/${projectId}`);
+  // const projectInfo = await universalGet<ProjectGetResponse>(`/api/projects/${projectId}`);
```
cguedes, line 31: Again :-) just to make sure you know all the changes.

src/api/rewrite.ts
```diff
@@ -14,7 +13,8 @@ export async function askForRewrite(selection: string, options?: RewriteOptions)
       ...DEFAULT_OPTIONS,
       ...options,
     };
-    const response = await universalPost<RewriteResponse, RewriteRequest>('/api/ai/rewrite', {
+    // const response = await universalPost<RewriteResponse, RewriteRequest>('/api/ai/rewrite', {
```
cguedes, line 16: Another one to remove.

src/api/typed-api.ts
```diff
@@ -0,0 +1,102 @@
+/** Type-safe access to the RefStudio HTTP APIs. */
```
cguedes, line 1: It would be amazing to have a blog post about this code and how typescript can help in such use-cases.

cguedes (dismissed):
This looks amazing 💪 
Only added some notes to remove commented code.

src/api/projectsAPI.ts
```diff
@@ -22,11 +23,15 @@ type ProjectPostResponse = ProjectsResponse;
...
+  const projectInfo = (await apiGetJson('/api/projects/{project_id}', {
+    path: { project_id: projectId },
+  })) as ProjectGetResponse;
```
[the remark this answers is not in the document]
danvk: Yeah, this is how that endpoint looks in `api-paths.ts`:

image

It comes back as `unknown`, so we need the type assertion.

src/api/typed-api.ts
danvk: Or a book chapter? :)

src/api/__tests__/typed-api.test.ts
```diff
@@ -0,0 +1,59 @@
...
+      // TODO: update request body type for this endpoint
+      {} as unknown as never,
+    );
+    expect(response).toEqual({
+      abc: { project_name: 'abc', project_path: '/path/to/project' },
+    });
+    expect(universalPost).toHaveBeenCalledWith('/api/projects/?project_name=abc', {});
+  });
+
+  it('should escape query parameters', async () => {
```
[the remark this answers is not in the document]
danvk: I have not. Is that something we anticipate needing? It looks like `URLSearchParameters` does support that.

danvk:
Oh, also, test coverage goes up because:

1. I'm excluding `api-types.ts`, which vitest previously counted as 285 uncovered lines.
2. `typed-api.ts` counts as 102 fully-covered lines.

Test coverage should probably ignore lines of code that only exist in type space (i.e. type declarations), but I'm not sure if vitest has a setting for that.

src/api/__tests__/typed-api.test.ts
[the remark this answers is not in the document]
cguedes: Not actually. But I think we will need. But agree that URLSearchParameters should be enough.

cguedes (approved):

PR #458, opened 2023-08-30
feat: Improve project details API

### LINKED ISSUE #448: Improve projects API and stop depending on Tauri FS types
The API output for the `/api/projects` (list projects) and `/api/projects/{id}` (project details) should improve to:

- not use full paths (the project path) bc it can be a security concern
- use a tree output for the project filesystem instead of a flat list of filepaths
- make sure we ignore know files (ex: .staging)
- adopt an output type compatible (close match) to `FileEntry.ts` instead of the `FileEntry` from `@tauri-apps/api/fs`

After these changes we should be able to update the client code to remove `convertTauriFileEntryToFileEntry`

This is related to #414 and a follow-up of #425.

gjreda:
backend fixes #448


python/sidecar/projects.py
```diff
@@ -106,7 +106,59 @@ def delete_project(user_id: str, project_id: str) -> None:
...
 
 
-def get_project_files(user_id: str, project_id: str) -> list:
+def get_project_files(user_id: str, project_id: str) -> dict:
+    """
+    Get the files in a project as a tree structure.
+
+    Returns
+    -------
+    dict
```
danvk, line 115: Could this be a pydantic structure instead of a `dict`?

gjreda:
@cguedes @sehyod I dropped the `is_file` and `is_folder` params y'all have on the TS side based on the convo in #448, but let me know if you'd like to keep them (see commit 776abd1)

python/tests/test_projects.py
```diff
@@ -175,16 +176,55 @@ def test_delete_project_should_delete_project(
...
+    project_name = projects.get_project_name(user_id, project_id)
+
+    # hidden dirs/files should be ignored
+    (Path(project_path) / "file1.refstudio").touch()
+    (Path(project_path) / ".ignoreme").touch()
+    project_path.joinpath("uploads").mkdir()
+    (Path(project_path) / "uploads" / "file2.pdf").touch()
+    project_path.joinpath(".storage").mkdir()
+    (Path(project_path) / ".storage" / "references.json").touch()
+    project_path.joinpath("empty").mkdir()
```
danvk, line 196: Could you add a nested directory here to cover that case?

python/tests/test_projects.py
```diff
@@ -175,16 +176,55 @@ def test_delete_project_should_delete_project(
...
+                    }
+                ],
+            },
+        ],
+    }
 
-    files = projects.get_project_files(user_id, project_id)
-    assert files == [Path(f"{project_path}/file1.txt")]
+    assert isinstance(response, ProjectFileTreeResponse)
+    assert sorted(response.dict()) == sorted(expected)
```
danvk, line 230: This doesn't quite work, `sorted(dict)` sorts whatever you'd get if you iterated over `dict`, which in this case is just the keys. So `sorted(expected) == ['contents', 'id', 'name', 'path']`.

I think the right way to handle this is to have `get_project_files` always return children in alphabetical order.

danvk (changes requested):
Could you merge in the latest changes on `main` so we could see how this looks in the new `api-paths.ts` file?

I had questions about how we want to handle nested directories and the test.

python/tests/test_projects.py
cguedes: We can handle sort in the client code. Maybe that's what makes most sense. CC @sehyod

python/sidecar/http.py
```diff
@@ -192,20 +193,13 @@ async def create_project(project_name: str, project_path: str = None):
 
 
 @project_api.get("/{project_id}")
-async def get_project(project_id: str):
+async def get_project(project_id: str) -> ProjectFileTreeResponse:
```
cguedes, line 196: This type can work for the /files endpoint, but don't work for the "project details" response.
Also, I think you should split getting the project files from the project information (id, name, ...).

I'm also I on dropping the project path completely from the output.
Note that in the future we might (will) have extra fields/props for a project like last updated date and alikes.

python/sidecar/http.py
```diff
@@ -230,10 +224,10 @@ async def delete_project(project_id: str):
 
 
 @project_api.get("/{project_id}/files")
-async def get_project_files(project_id: str):
+async def get_project_files(project_id: str) -> ProjectFileTreeResponse:
```
cguedes, line 227: This return type is ok, but only with the project files and not the id, name, ...

cguedes (changes requested):

python/sidecar/projects.py
```diff
@@ -106,7 +107,62 @@ def delete_project(user_id: str, project_id: str) -> None:
...
+        relpath = obj.relative_to(project_path)
+
+        # Skip hidden files and directories
+        if relpath.name.startswith(".") or obj.name.startswith("."):
+            continue
+
+        # files in subdirectories
+        if obj.is_dir():
+            contents.append(create_folder_entry(obj, project_path=project_path))
+            continue
```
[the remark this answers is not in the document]
gjreda: Had to google what that trap was. Wouldn't this be the opposite of that though since `continue` jumps to the next item in the iteration?

I can change though.

python/tests/test_projects.py
gjreda: Agree that client side handling sort makes the most sense.

python/sidecar/projects.py
```diff
@@ -106,7 +107,62 @@ def delete_project(user_id: str, project_id: str) -> None:
...
+    contents = []
+    for obj in project_path.rglob("*"):
+        relpath = obj.relative_to(project_path)
+
+        # Skip hidden files and directories
+        if relpath.name.startswith(".") or obj.name.startswith("."):
+            continue
+
+        # files in subdirectories
+        if obj.is_dir():
```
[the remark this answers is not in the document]
gjreda: > We can drop isFile and use the undefined children to detect if the entry is a file or folder.

@cguedes can you clarify that piece from the issue?

With the current code, this would be the structure for a nested directory:
```
[{
	"name": "empty",
	"path": "empty",
	"children": []
}, {
	"name": "uploads",
	"path": "uploads",
	"children": [{
		"name": "file2.pdf",
		"path": "uploads/file2.pdf",
		"file_extension": ".pdf"
	}]
}, {
	"name": "nested",
	"path": "nested",
	"children": [{
		"name": "file3.pdf",
		"path": "nested/folder/file3.pdf",
		"file_extension": ".pdf"
	}]
}, {
	"name": "file1.refstudio",
	"path": "file1.refstudio",
	"file_extension": ".refstudio"
}, {
	"name": "folder",
	"path": "nested/folder",
	"children": [{
		"name": "file3.pdf",
		"path": "nested/folder/file3.pdf",
		"file_extension": ".pdf"
	}]
}]
```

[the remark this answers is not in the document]
cguedes: No I would expect a tree of **File** / **Directory** entries.

For this scenario it would be:

Note the `folder` directory is sibling of `file3.pdf` and that `file4.pdf` is "inside" the `folder` as children.


```json
[
  {
    "name": "empty",
    "path": "empty",
    "children": []
  },
  {
    "name": "uploads",
    "path": "uploads",
    "children": [
      {
        "name": "file2.pdf",
        "path": "uploads/file2.pdf",
        "file_extension": ".pdf"
      }
    ]
  },
  {
    "name": "nested",
    "path": "nested",
    "children": [
      {
        "name": "file3.pdf",
        "path": "nested/file3.pdf",
        "file_extension": ".pdf"
      },
      {
        "name": "folder",
        "path": "nested/folder",
        "children": [
          {
            "name": "file4.pdf",
            "path": "nested/folder/file4.pdf",
            "file_extension": ".pdf"
          }
        ]
      }
    ]
  },
  {
    "name": "file1.refstudio",
    "path": "file1.refstudio",
    "file_extension": ".refstudio"
  }
]
```

python/sidecar/http.py
```diff
@@ -192,20 +195,13 @@ async def create_project(project_name: str, project_path: str = None):
...
-    project_path = projects.get_project_path(user_id, project_id)
-    project_name = projects.get_project_name(user_id, project_id)
-    filepaths = projects.get_project_files(user_id, project_id)
-    return {
-        "project_id": project_id,
-        "project_path": project_path,
-        "project_name": project_name,
-        "filepaths": filepaths,
-    }
+    response = projects.get_project_files(user_id, project_id)
```
cguedes, line 203: why do we need to call `get_project_files` to only retrieve the project details? No need to read and create the filesystem structure for this endpoint.

Also, the docs for this method are wrong.

python/sidecar/http.py
```diff
@@ -230,10 +226,10 @@ async def delete_project(project_id: str):
 
 
 @project_api.get("/{project_id}/files")
-async def get_project_files(project_id: str):
+async def get_project_files(project_id: str) -> list[FileEntry | FolderEntry]:
```
cguedes, line 229: nit: I would prefer to have a type (ex: `ProjectFiles`) that would contain a filed with this list. Returning arrays in the output prevents the output from evolving.

python/sidecar/http.py
```diff
@@ -230,10 +226,10 @@ async def delete_project(project_id: str):
...
     user_id = "user1"
-    filepaths = projects.get_project_files(user_id, project_id)
-    return filepaths
+    response = projects.get_project_files(user_id, project_id)
```
cguedes, line 231: IMO this method should only return the filesystem content.

python/sidecar/projects.py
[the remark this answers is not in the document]
cguedes: The output is still wrong 
image

python/sidecar/typing.py
```diff
@@ -206,6 +206,30 @@ class SearchResponse(RefStudioModel):
...
+    name: str
+    path: str
+
+
+class FileEntry(FileEntryBase):
+    file_extension: str
+
+
+class FolderEntry(FileEntryBase):
+    children: list[FileEntry]
```
cguedes, line 219: This should `children: list[FileEntry | FolderEntry]`

python/sidecar/typing.py
```diff
@@ -206,6 +206,30 @@ class SearchResponse(RefStudioModel):
...
+
+
+class ProjectDetails(RefStudioModel):
+    id: str
+    name: str
+    path: str
+
+
+class ProjectFileTreeResponse(RefStudioModel):
```
cguedes, line 228: nit: I think that we only use *Response types for the API output. This type is only used now in `projects.py`

python/sidecar/typing.py
```diff
@@ -206,6 +206,30 @@ class SearchResponse(RefStudioModel):
...
+
+class FileEntry(FileEntryBase):
+    file_extension: str
+
+
+class FolderEntry(FileEntryBase):
+    children: list[FileEntry]
+
+
+class ProjectDetails(RefStudioModel):
```
cguedes, line 222: nit: we usually call the API output types `....Response`

cguedes (changes requested):

python/sidecar/typing.py
gjreda: We _usually_ do, but we do not always. I'm finding repeatedly using `Response` to be redundant in my cases. If a function returns a type, it is by definition a "response."

python/sidecar/http.py
gjreda: that's what's happening on the line below

python/sidecar/typing.py
cguedes: You are right, we use sometimes.

I find it useful to have a naming convention for the API output types in order to distinguish them to internal types. That also help the client code to use the correct type.

python/sidecar/http.py
cguedes: I see that. My concern is why do you call `projects.get_project_files(user_id, project_id)` if you don't need the files. That's extra work (reading the filesystem) for nothing. I think you should create methods to split that method (of projects) in two (one to read project info and the other for files)

cguedes (approved):

gjreda:
That dismissal sounds more curt than I intended. I also addressed @danvk's comments in my changes 😅

PR #463, opened 2023-08-31
feat: update styles of rewrite panel

### LINKED ISSUE #375: Design for Rewriter Panel

sehyod:
This updates the styles of the Rewrite Panel

https://github.com/refstudio/refstudio/assets/58954208/b72953ef-ad8f-4f50-ae77-397efe2f8ca9

Closes #375


cguedes (changes requested):
Some notes:

- We should keep the heading "Pick Variation ..." fixed and only scroll the choice
- I did expected an animation for when switching between variations
image
image


- The sider bg %age is incorrectly computed for this scenario
image


- we should have a better text for this popup. Ask Jeff in slack.
image


- Create a ticket to implement the rewrite inside the command palette
image

sehyod:
Thank you for your comments.
- I have pushed the fix to make the "Pick Variation" fixed
- I have asked the design team if they think we should add an animation there
- I will look into this issue
- I will ask about the copy, that's why I marked it as TODO in the code
- The ticket is #471

sehyod:
I have updated the tooltip and the design team confirmed there should not be an animation when switching between variations @cguedes

cguedes (approved):

PR #465, opened 2023-08-31
Create distinct types for Settings / SettingsPatch

danvk:
This solves the problem with optional fields in `SettingsSchema` in TypeScript.

In general I think we should avoid specifying default values in Pydantic models that are part of the HTTP API because the semantics are unclear. If you have something like:

```py
class Settings(RefStudioModel):
    a: str = 'alice'
    b: str = 'bob'

@settings_api.put("/")
async def update_settings(req: SettingsSchema) -> SettingsSchema:
    ...
```

Then if the users hits the PUT endpoint with `{"a": "archer"}`, presumably their intention is to just set "a" to "archer" and not also set "b" to "bob".

Highlights:

- I "flattened" the settings schema like we talked about last week. The new model is `FlatSettingsSchema` and it has no `Optional` fields or fields with default values.
- I found [this decorator][1] to be the best way to make an all-optional version of a Pydantic model. This is like `Partial<T>` in TypeScript. I used this to define a `FlatSettingsSchemaPatch` model. This might be a useful pattern going forward. But if the update type is something more complicated (say you want to drop an `id` field that can't be updated) then maybe just writing out the two models explicitly is the way to go.
- Since the `settings.json` format has changed, I wrote a migration function.
- With a flat settings structure, the `getDotNotation` from `tauri-settings` isn't useful anymore and we can replace it with something much simpler. In fact, we can get rid of `tauri-settings` altogether! 🎉 

[1]: https://stackoverflow.com/a/74296358/388951

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,6 @@
     "react-popper": "^2.3.0",
     "react-resizable-panels": "^0.0.51",
     "tailwindcss": "^3.3.2",
-    "tauri-settings": "^0.3.2",
     "tiptap-markdown": "^0.7.2",
     "turndown": "^7.1.2",
     "usehooks-ts": "^2.9.1"
--- a/python/sidecar/settings.py
+++ b/python/sidecar/settings.py
@@ -1,11 +1,16 @@
+import json
 import logging
 import os
-import json
 from pathlib import Path
 
 from dotenv import load_dotenv
-
 from sidecar.typing import SettingsSchema
+
+from python.sidecar.typing import (
+    FlatSettingsSchema,
+    FlatSettingsSchemaPatch,
+    RewriteMannerType,
+)
 
 load_dotenv()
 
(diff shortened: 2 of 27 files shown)


python/sidecar/settings.py
```diff
@@ -35,17 +39,46 @@ def make_settings_json_path(user_id: str) -> Path:
     return filepath
 
 
+def default_settings() -> FlatSettingsSchema:
```
danvk, line 42: This is a way to define default settings without adding default values to the type.

python/sidecar/typing.py
```diff
@@ -230,11 +231,28 @@ class SidecarSettings(RefStudioModel):
...
+    current_directory: str
+    logging_enabled: bool
+    logging_filepath: str
+    openai_api_key: str
+    openai_chat_model: str
+    openai_manner: RewriteMannerType
+    openai_temperature: float
+
+
+@make_optional()
```
danvk, line 251: This is how you make an optional model with the decorator

src/api/api-paths.ts
```diff
@@ -198,6 +195,26 @@ export interface components {
       message: string;
       status: ResponseStatus;
     };
+    /** FlatSettingsSchema */
+    FlatSettingsSchema: {
```
danvk, line 199: The new types are much nicer -- all fields required!

src/settings/settingsManager.ts
```diff
@@ -1,54 +1,50 @@
...
-import { apiGetJson } from '../api/typed-api';
-
-export type OpenAiManner = apiTypes.RewriteMannerType;
-export function getMannerOptions(): OpenAiManner[] {
+export function getMannerOptions(): RewriteMannerType[] {
   return ['concise', 'elaborate', 'scholarly'];
 }
 
-// TODO: remove this after we separate the request/response types in the settings API
-type DeepRequired<T> = T extends object ? { [k in keyof T]-?: DeepRequired<T[k]> } : T;
```
danvk, line 15: ✂️ 🍾 🎉

src/settings/settingsManager.ts
```diff
@@ -1,54 +1,50 @@
...
+}
 
-let settingsManager: SettingsManagerView | undefined;
+let settingsManager: SettingsManagerView<FlatSettingsSchema> | undefined;
 
 export async function initSettings() {
-  let settings: SettingsSchema;
+  let settings: FlatSettingsSchema;
   try {
-    settings = (await apiGetJson('/api/settings/')) as SettingsSchema;
```
danvk, line 33: no more type assertion! 🎉

package.json
```diff
@@ -74,7 +74,6 @@
     "react-popper": "^2.3.0",
     "react-resizable-panels": "^0.0.51",
     "tailwindcss": "^3.3.2",
-    "tauri-settings": "^0.3.2",
```
cguedes, line 77: 👏 🎉

python/sidecar/settings.py
```diff
@@ -35,17 +39,46 @@ def make_settings_json_path(user_id: str) -> Path:
...
 
-    defaults = SettingsSchema()
+    defaults = default_settings()
 
     with open(filepath, "w") as f:
         json.dump(defaults.dict(), f)
 
 
-def get_settings_for_user(user_id: str) -> SettingsSchema:
+def migrate_settings(old_settings: dict) -> FlatSettingsSchema:
```
cguedes, line 64: wondering how/when we will get rid of this migration (fallback) code.

cguedes (approved):
Only added a note about the migrate_settings utility

python/sidecar/settings.py
danvk: Maybe in a week? Or since we don't care about backwards compatibility yet, you could be bold and take it out now :)

PR #466, opened 2023-08-31
Port a few more endpoints to the typed API

danvk:
I filled in the FastAPI / Pydantic schemas in a few places. Note that POST requests need a JSON payload with this system. If we want to support empty payloads, we'll need to add some more logic in `typed-api.ts`.

### DIFF
--- a/python/sidecar/http.py
+++ b/python/sidecar/http.py
@@ -13,11 +13,11 @@ from sidecar.typing import (
     ChatResponse,
     DeleteRequest,
     DeleteStatusResponse,
+    EmptyRequest,
     IngestRequest,
     IngestResponse,
     Reference,
     ReferencePatch,
-    RefStudioModel,
     RewriteRequest,
     RewriteResponse,
     SearchRequest,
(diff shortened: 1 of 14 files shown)


cguedes:
> Note that POST requests need a JSON payload with this system. 

I agree that POST should alway have a JSON payload, even though sometimes it can be empty.

cguedes (approved):
Approved with a note about the create project payload (querystring vs post body). We should update that later.

sehyod (approved):

python/sidecar/http.py
```diff
@@ -166,7 +168,9 @@ async def list_projects():
 
 
 @project_api.post("/")
-async def create_project(project_name: str, project_path: str = None):
+async def create_project(
+    payload: EmptyRequest, project_name: str, project_path: str = None
```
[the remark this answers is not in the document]
cguedes: I did this change when I merged the from main.

cguedes (approved):

python/sidecar/http.py
```diff
@@ -334,3 +328,4 @@ async def get_settings() -> FlatSettingsSchema:
 async def update_settings(req: FlatSettingsSchemaPatch) -> FlatSettingsSchema:
     user_id = "user1"
     return settings.update_settings_for_user(user_id, req)
+    return settings.update_settings_for_user(user_id, req)
```
gjreda, line 331: This shouldn't have been merged

cguedes: Nice catch 😅

danvk: I've noticed this happen on a few of our Python files. I think something that runs when we save Python files must be doing it. Total shot in the dark but maybe it has something to do with missing trailing newlines?

Issue #414, opened 2023-08-24
Improve the projects API (output payload and types)


cguedes:
This is a follow-up after https://github.com/refstudio/refstudio/pull/393#pullrequestreview-1593894064

We should also make sure we can configure a project name.

CC @danvk

danvk:
FastAPI generates OpenAPI JSONSchema here: https://fastapi.tiangolo.com/tr/tutorial/first-steps/#openapi

You can see it for our various APIs here:
http://127.0.0.1:8000/api/sidecar/openapi.json

It can be generated programmatically using this recipe: https://github.com/tiangolo/fastapi/issues/1173#issuecomment-605664503

gjreda:
@cguedes I think this done now?

cguedes:
Yes, this is done.

PR #472, opened 2023-09-01
feat: save as markdown behaviour for web

### LINKED ISSUE #455: (discussion) Save as Markdown behaviour for web
In the desktop, the action "Save as Markdown" (via menu) uses a Tauri dialog to choose a target folder that defaults to folder inside the project, but the user can select a different folder in the OS filesystem. That folder will have a .md file and optionally a .bib file with reference(s) used in the document.

For the web, we have two issues:
 1. Where should we have the action to trigger the action?
 2. What should be the UX for the action.

My considerations:
1. Can be implemented in the command palette and also as a context context menu action in the files explorer.
2. We don't need an alternative to the Tauri dialog, saving the export in an export folder inside the project (maybe under a folder named after the file we want to export so that we can have multiple files / exports). Then we can implement a download feature of files/folder to access these files from web.
   a. Should we implement the same behaviour for the desktop version?

----

This is a follow-up after #425

cguedes:
This PR fixes #455 by implementing the same codebase for desktop and web, saving the files (markdown, and optionally the .bib) in the exports folder inside the project.

We can't add the operation to the explorer context menu because we can't ensure that the entry is hidden/inactive for other file types (pdf, .md), and also the current codebase exports the active editor (has a dependency on the TipTap editor instance)

image

image


src/features/textEditor/components/TipTapEditor.tsx
```diff
@@ -102,3 +108,26 @@ export function TipTapEditor({ editorContent, editorId, isActive, saveFileInMemo
     </div>
   );
 }
+
+export async function saveAsMarkdown(markdownSerializer: MarkdownSerializer, filePath: string) {
```
cguedes, line 112: I've moved this file here as he depends in the markdownSerializer that belong to the "text editor".
I've also simplified the code by using the new `getFileNameAndExtension`.

src/api/projectsAPI.ts
```diff
@@ -81,14 +85,31 @@ export async function readProjectFiles(projectId: string): Promise<TauriFileEntr
       children: undefined,
     }));
 
+  const exportsFiles = relativePaths
```
cguedes, line 88: I've added this support while we don't have #414 ready.

src/application/App.tsx
```diff
@@ -43,7 +43,7 @@ export function App() {
   );
 
   useEventListener('unhandledrejection', (event) => {
-    notifyError('An error occurred', `Unhandled rejection in promise. Reason: ${event.reason}).`);
+    notifyError('Unhandled rejection in promise', `Reason: ${event.reason}`);
```
cguedes, line 46: Improve this a bit further.

src/features/textEditor/components/TipTapEditor.tsx
```diff
@@ -81,8 +84,11 @@ export function TipTapEditor({ editorContent, editorId, isActive, saveFileInMemo
     const mdSerializer = new MarkdownSerializer(editor, references);
     const { id: filePath } = parseEditorId(editorId);
 
-    void saveAsMarkdown(mdSerializer, filePath);
-  }, [editor, editorId, isActive, references]);
+    void saveAsMarkdown(mdSerializer, filePath).then(() => {
+      // We need to make sure the files were saved before refreshing the file tree.
+      void refreshFileTree();
```
cguedes, line 89: This was missing.

src/features/textEditor/components/TipTapEditor.tsx
cguedes: After our chat I moved this to own file in `src/features/textEditor`

src/features/textEditor/components/TipTapEditor.tsx
sehyod: Nice thank you for that!
You can simplify it like this:
```typescript
void saveAsMarkdown(mdSerializer, filePath)
      // We need to make sure the files were saved before refreshing the file tree.
      .then(refreshFileTree);
```

src/io/filesystem.ts
```diff
@@ -140,6 +146,15 @@ export function splitRefStudioPath(filePath: string): string[] {
...
 }
 
+export function getFileNameAndExtension(filePath: string): { name: string; ext: string } {
+  const fileName = filePath.slice(filePath.lastIndexOf(sep) + 1);
+  const dotIndex = fileName.lastIndexOf('.');
+  if (dotIndex === -1) {
+    return { name: fileName, ext: '' };
+  }
+  return { name: fileName.slice(0, dotIndex), ext: fileName.slice(dotIndex + 1) };
+}
```
sehyod, line 156: There is an issue with dot files: for a `.gitignore` file, the function would return `{ name: '', extension: 'gitignore' }`
Even if we don't have dot files for now, I think it would be safer to treat this edge case

sehyod (changes requested):

sehyod (approved):

PR #476, opened 2023-09-01
feat: Add active project id to settings

### LINKED ISSUE #445: Update settings to adopt project ID instead of project path
We should update this code in `projectEventListeners.ts` to use a new key.

```js
// TODO: This method needs to be refactored to have use a project id instead of a path
function persistProjectIdInSettings(id: string) {
  setCachedSetting('project.current_directory', id);
  void saveCachedSettings();
}
```

Tasks:
- [x] Add `active_project_id` to the settings backend API
- [x] Adopt the new `active_project_id` in the client code
- [x] #494
- [x] #504

No setting migration is needed.

This is a follow-up task after #425

gjreda:
fixes #445

### DIFF
--- a/python/api.schema.json
+++ b/python/api.schema.json
@@ -640,6 +640,9 @@
     },
     "FlatSettingsSchema": {
       "properties": {
+        "active_project_id": {
+          "type": "string"
+        },
         "current_directory": {
           "type": "string"
         },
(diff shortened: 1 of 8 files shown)


gjreda:
cc @cguedes

cguedes (approved):

PR #479, opened 2023-09-04
feat: Implement design for settings modal

sehyod:
This implements useful components (Checkbox, Input, ...) and the updated design for the Settings Modal.
The modal can now be closed by clicking outside of it or pressing Escape

Screenshot 2023-09-04 at 14 13 55

Closes #438

Issue #447, opened 2023-08-30
Open Project Dialog for Web


cguedes:
In web, the app reads the first project returned by the projects API instead of listing all projects and open the one selected by the user.

This is a follow-up after https://github.com/refstudio/refstudio/pull/425

PR #480, opened 2023-09-04
feat: new project dialog for web

### LINKED ISSUE #446: New Project Dialog for Web
In web, the app uses the hard-coded "Project Web" project name instead of asking the user for a name.

This is a follow-up after #425

cguedes:
This closes #446 

This PR introduces a new core atom for modals named `createModalAtoms`.

With that, we can create a modal atoms for the create project model with:

```ts
export const createProjectModalAtoms = createModalAtoms<string>();
```

And then open the modal using:

```ts
const modalResult = useSetAtom(createProjectModalAtoms.openAtom);
if (modalResult.status === 'dismissed') {
   // modal closed or canceled
}
const projectName = modalResult.value;
...
```

That would return a string with the modal return value (a string), or `null` if the modal was closed/canceled.

The modal code would be something like:

```ts
  const isVisible = useAtomValue(createProjectModalAtoms.visibleAtom);
  const closeModal = useSetAtom(createProjectModalAtoms.closeAtom);
  const dismissModal = useSetAtom(createProjectModalAtoms.dismissAtom);
  ...
  const handleFormSubmit: FormEventHandler = (e) => {
    ...
    closeModal(name.trim());
  };
  const handleOnCancel: MouseEventHandler<HTMLButtonElement> = (e) => {
    dismissModal();
  };

```

image

image


src/application/CreateProjectModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+    closeModal(null);
+  };
+
+  if (!isVisible) {
+    return <></>;
+  }
+
+  return (
+    <div className="fixed left-0 top-0 z-modals flex h-screen w-screen items-center justify-center">
+      <div className="pointer-events-none absolute left-0 top-0 h-full w-full bg-slate-400/60" />
```
sehyod, line 37: A generic Modal component will be added in #479. Do you want to wait for it before merging this PR?

src/components/Button.tsx
```diff
@@ -42,7 +46,8 @@ export function Button({
         className,
       )}
       disabled={disabled}
-      onClick={disabled ? undefined : onClick}
+      type={submit ? 'submit' : 'button'}
+      onClick={disabled || !onClick ? undefined : onClick}
```
sehyod, line 50: I don't think `|| !onClick` is necessary here

src/application/CreateProjectModal.tsx
cguedes: I will.

src/components/Button.tsx
cguedes: Sure!

sehyod (approved):

src/application/CreateProjectModal.tsx
```diff
@@ -1,65 +1,44 @@
...
-      <div className="pointer-events-none absolute left-0 top-0 h-full w-full bg-slate-400/60" />
-      <div
-        className={cx(
-          'relative w-[1150px] max-w-[calc(30vw-100px)] overflow-hidden',
-          'rounded-xl border-none border-slate-500 bg-white shadow-xl shadow-slate-500',
-          '',
-        )}
-      >
+    <Modal className="w-[1150px] max-w-[calc(30vw-100px)]" open={isVisible} onClose={() => dismissModal()}>
+      <div className="flex w-full flex-col  bg-white">
```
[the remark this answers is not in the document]
cguedes: Great catch. Will fix that in the next issue for #447

PR #488, opened 2023-09-05
fix: margin between sibling list items

### LINKED ISSUE #481: Vertical spacing for lists is off
Here's an example of indenting an ordered list item. You can see that, upon indent, the item moves down, and then moves back up when un-indented.

https://github.com/refstudio/refstudio/assets/15233/1a3f23cf-0bb8-4bfa-ba42-5104d2415e10

This doesn't happen with Notion. The item stays level:

https://github.com/refstudio/refstudio/assets/15233/f565ed3b-7638-406e-a65e-4aceb1d0b213

Other vertical spacing weirdness includes converting collapsible blocks to an unordered list (one of the collapsible list items expands out of the recording window):

https://github.com/refstudio/refstudio/assets/15233/dd48986b-411d-4e64-9cac-9f2972aadbb8

Again not a problem on Notion:

https://github.com/refstudio/refstudio/assets/15233/8205b30c-5339-4085-96ab-641e39f8c67c

sehyod:
There were 2 issues:
 - The spacing between list items was 0 instead of being the same as the spacing between a parent node and its child, resulting in a weird shifting behaviour when indenting/un-indenting an item in a list
 - The spacing between list items was different from the spacing between collapsible nodes.

The first issue was a bug and has been fixed. The second one was more or less intended: the design specifies that there should be 24px between each node:

image

I have changed that to make collapsible nodes considered as list items (which makes sense because that is how we export them in markdown). However, the default spacing remains 24px, which means that turning a regular node into a collapsible node still results in things moving up and down:

https://github.com/refstudio/refstudio/assets/58954208/3d1fb8b3-7379-40f3-b8aa-b4db6c4cf89e

On notion, the spacing is the same between every kind of node, except headings. So the question is do we want to do the same and remove this 24px spacing we have @hammer?

Fixes #481

### DIFF
--- a/src/features/textEditor/components/tipTapNodes/notionBlock/NotionBlock.css
+++ b/src/features/textEditor/components/tipTapNodes/notionBlock/NotionBlock.css
@@ -111,15 +111,10 @@
 }
 
 .node-notionBlock + .node-notionBlock {
-  @apply mt-6;
+  @apply mt-2;
 }
 
 /* Add space between a parent notion block and its first child */
 .node-notionBlock .node-notionBlock:first-of-type {
-  @apply mt-2;
-}
-
-/* Space between sibling list items */
-.node-notionBlock:has(> * > .list-item) + .node-notionBlock:has(> * > .list-item) {
   @apply mt-2;
 }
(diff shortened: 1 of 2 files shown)


hammer:
> On notion, the spacing is the same between every kind of node, except headings. So the question is do we want to do the same and remove this 24px spacing we have @hammer?

Yes please! The Notion behavior is more natural to me.

cguedes (approved):
👍

PR #495, opened 2023-09-05
Reorg FastAPI backend into submodules

### LINKED ISSUE #415: Break up http.py

### LINKED ISSUE #494: Remove the deprecated `current_directory` setting from the backend API

gjreda:
fixes #415

Break up the backend into submodules. This layout largely follows the [FastAPI best practices listed here](https://github.com/zhanymkanov/fastapi-best-practices), which is inspired by Netflix's [Dispatch](https://github.com/Netflix/dispatch/tree/master) project.

--------
[fc10954](https://github.com/refstudio/refstudio/pull/495/commits/fc10954ebd29df2e5e2fd80dc7ab2d9cf1189dcb) also fixes #494. Threw it on here since it was going to lead to merge conflicts.


cguedes:
@gjreda what do you think of removing the top level `sidecar` folder (well, keep it to the server cmd)? IMO That's now more of an implementation detail.

gjreda:
@cguedes I'd like to keep it (or rename it) in order to follow a [fairly standard python project layout](https://python-poetry.org/docs/basic-usage/#project-setup). `sidecar` still feels like a fairly apt name (as does `backend` or something similar).

cguedes:
@gjreda two notes before the review:

- we need to update the `project.json` script for the `yarn web:api` 
- I would like (need) the API url to start with a `/api`. This is useful to create a (rewrite/proxy) rule when running the client in dev mode. Would also be useful if we serve the client app (html, css, ...) via the same backend server.

gjreda:
@cguedes I think this is good to go now. The codegen schemas look ok to me, but can you verify on your end as well? There aren't any tests for that code, so it was hard to verify that the changes were ok.

I'll write some tests for `generate_schema.py` in a follow-up PR.

cguedes (approved):

PR #500, opened 2023-09-06
feat: add empty welcome view

### LINKED ISSUE #499: Add empty welcome screen view

sehyod:
This implements the empty view for the welcome screen (when the user has no project)
This loads the list of projects in an atom during app startup, so that the list is already available when displaying the welcome view

image

Closes #499

### DIFF
--- a/src/application/components/ProjectsList.tsx
+++ b/src/application/components/ProjectsList.tsx
@@ -6,7 +6,7 @@ import { RefStudioEditorIcon } from './icons';
 
 interface ProjectsListProps {
   projects: ProjectInfo[];
-  onProjectClick: (projectId: ProjectInfo) => void;
+  onProjectClick: (project: ProjectInfo) => void;
 }
 export function ProjectsList({ projects, onProjectClick }: ProjectsListProps) {
   return (
--- a/src/application/components/ProjectsList.tsx
+++ b/src/application/components/ProjectsList.tsx
@@ -31,6 +31,7 @@ function ProjectItem({ project, onClick }: { project: ProjectInfo; onClick: () =
         'rounded-default hover:bg-btn-bg-side-bar-item-hover',
         'text-btn-txt-side-bar-item-primary',
       )}
+      role="menuitem"
       onClick={onClick}
     >
       <div className="text-btn-ico-content">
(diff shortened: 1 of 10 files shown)


src/application/views/WelcomeView.tsx
```diff
@@ -1,30 +1,60 @@
...
 import { OpenIcon } from '../../components/icons';
 import { emitEvent } from '../../events';
 import { cx } from '../../lib/cx';
-import { AddIcon, RefStudioEditorIcon, SampleIcon } from '../components/icons';
+import { AddIcon, EmptyStateIcon, SampleIcon } from '../components/icons';
+import { ProjectsList } from '../components/ProjectsList';
 
 export function WelcomeView() {
+  const projects = useAtomValue(allProjectsAtom);
+
```
cguedes, line 13: How is the list of projects (atom) updated?
I think we should use an effect to update the atom one we land on this view. What do you think?

src/application/components/ProjectsList.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+function ProjectItem({ project }: { project: ProjectInfo }) {
+  return (
+    <div
+      className={cx(
+        'flex cursor-pointer items-center gap-2 p-2 pr-3',
+        'rounded-default hover:bg-btn-bg-side-bar-item-hover',
+        'text-btn-txt-side-bar-item-primary',
+      )}
+      onClick={() => emitEvent('refstudio://projects/open', { projectId: project.id })}
```
cguedes, line 31: Can you add an event to this component (and the parent) for clicks in the project, and then implement the emitEvent outside. This will allow me to reuse this in #447

cguedes (changes requested):
Added two notes about refresh of the projects atom and for the ProjectsList component

src/application/views/WelcomeView.tsx
sehyod: You're right! Good catch!
I'm wondering whether the update should be done when creating a project instead? The issue with updating it here is that we would first render the outdated list, then new projects would appear once the list is updated and that could look weird because there would not be any loading effect

cguedes: I've also thought about that. My concern was that the list can be updated externally. For now your idea is a good solution.

cguedes (approved):

PR #502, opened 2023-09-06
fix: prevent panel from uncollapsing when sidebar is disabled

### LINKED ISSUE #486: Explorer open when no project is open
Sometimes the app starts with not project open but the file explorer open, although it should be disabled

sehyod:
**Problem**
Resizable panels' layout is stored in local storage. When starting the app with sidebars uncollapsed, the library will force them to be uncollapsed, even if the sidebar should be disabled.

Screenshot 2023-09-06 at 12 42 30

**Solution**
Prevent the automatic uncollapse action performed by the library when the sidebar is disabled.

Screenshot 2023-09-06 at 12 43 33

Fixes #486


cguedes (approved):
🎉

PR #508, opened 2023-09-06
Add workflow to create Python binaries and build Refstudio for supported OS

### LINKED ISSUE #334: Error: `Bad CPU type in executable (os error 86)` when trying to use AI Selection feature
When I was testing the released application I encountered an error, specifically when using the `Selection` feature. I selected some text and then requested the app to rewrite it but I got an error saying `Bad CPU type in executable (os error 86)`.

image

I suspected it could be the version I was executing on my machine (Apple chip version on a Intel computer) but I checked the `.dmg` and it was the Intel version. Then I tried with the `universal` one, but unfortunately got the same error.

> **Note**: This doesn't happen on boot, only when using the `Selection` feature

For reference, this are the specs of my machine:
- OS Version: Ventura 13.4.1 (c)
- Chip: Intel Core i7

gjreda:
fixes #334

Adds a GH action (reuse cut-release.yml) to build refstudio for both mac architectures, and release (as draft) to github.
The release version is read from the `package.json`. For now we added the next version (23.2.0) with an `-alpha.1` suffix.

This PR also fixes an issue with the react-pdf in intel chips. We've had to add polyfill and use the legacy pdf worker.
Additionally add some extra startup logs (main window -> splashscreen) to debug.

_____________

**macOS**
~Almost there via [1c991af](https://github.com/refstudio/refstudio/pull/508/commits/1c991aff6d6658a562d3a076883687681513f0a9)~

- [x] Create Python binary
- [x] Verify the Python binary works (start server in CI process and `curl` status endpoint)
- [x] Tauri builds successfully
- [x] Release is created with tauri builds attached
- [x] Refstudio installs
- [x] Sidecar python binary works when launched from install location `/Applications/refstudio.app/Contents/Resources/bin/python/main`
- [x] Refstudio works

------

The following issue with linux will be follow-up in #540 

Current status:
**Linux**
Python binary builds, but tauri build process fails ([workflow](https://github.com/refstudio/refstudio/actions/runs/6130704756/job/16640045919))
```
ERROR: Could not find dependency: libquadmath-96973f99.so.0.0.0 
ERROR: Failed to deploy dependencies for existing files 
       Error [tauri_cli_node] failed to bundle project: error running appimage.sh
Error: Command failed with exit code 1: yarn tauri build --verbose --target x86_64-unknown-linux-gnu
```

PS: Edited by @cguedes at Sep 13.


hammer:
Nice! If I understand correctly, this will build for macOS x86 (Intel) and Windows.

I guess we will still need to manually build for macOS ARM (Apple Silicon)?

The beta for GitHub-hosted Apple Silicon does not yet appear to be open: https://github.com/github/roadmap/issues/528

Or we can try to use the self-hosted runner (https://github.blog/changelog/2022-08-09-github-actions-self-hosted-runners-now-support-apple-m1-hardware) with the AWS EC2 M1 instances (https://aws.amazon.com/about-aws/whats-new/2022/07/general-availability-amazon-ec2-m1-mac-instances-macos/)? This repo looks to have some code for that: https://github.com/machulav/ec2-github-runner.

hammer:
Also should we throw a `ubuntu-latest` on there in case it just works for Linux too?

gjreda:
@hammer Makes sense. My goal here is to find a workflow that works for one OS and then layer in more.

hammer:
You might be able to test the Ubuntu build with the [AWS free tier](https://aws.amazon.com/free/?all-free-tier.sort-by=item.additionalFields.SortRank&all-free-tier.sort-order=asc&awsf.Free%20Tier%20Types=*all&awsf.Free%20Tier%20Categories=*all), though it might be a pain to fiddle with the firewall.

hammer:
[UTM](https://mac.getutm.app/) looks like another option for testing other OS builds on Mac.

.github/workflows/build.yml
```diff
@@ -0,0 +1,116 @@
...
+  #           })
+  #           return data.id
+    
+  build-tauri:
+    # needs: create-release
+    permissions:
+      contents: write
+    strategy:
+      fail-fast: false
+      matrix:
```
sehyod, line 45: Now we have an os matrix, you need to change the cache key for poetry `key: poetry-cache-${{ runner.os }}-${{ hashFiles('.github/actions/setup-python/action.yml') }}` (in `.github/actions/setup-python/action.yml`, line 15)

cguedes:
The root cause for the build/image issue for the Apple Intel Chip is the [PDF library](https://github.com/wojtekmaj/react-pdf/) we use to display PDFs.
I deactivated that library (PDFView) and now the application starts without issues.

image

image

hammer:
> The root cause for the build/image issue for the Apple Intel Chip is the [PDF library](https://github.com/wojtekmaj/react-pdf/) we use to display PDFs.

Weird! I searched their issue tracker for “intel” and “x86” but didn’t find anything. Should we report this upstream? Do you have any idea what might be causing it?

hammer:
> @cguedes @hammer this issue in their repo looks related: [wojtekmaj/react-pdf#1465](https://github.com/wojtekmaj/react-pdf/issues/1465)

🤔 so it’s a browser issue, not an OS issue? We should probably add a browser matrix to our tests and maybe builds, sigh…

cguedes:
> Weird! I searched their issue tracker for “intel” and “x86” but didn’t find anything. 
@hammer I also haven't found anything. 

> @cguedes @hammer this issue in their repo looks related: https://github.com/wojtekmaj/react-pdf/issues/1465
Nice catch. I'm investigating this.

Meanwhile I'm also testing a new approach without a lib for showing PDFs, using the native browser PDF preview (via iframe). This can be implemented now as we have a URL for the PDF using the filesystem API. 

The cons of this solution are:
1. Can't interact with the PDF content (is an `<embed .... />`), we can only configure [how the viewer looks](https://tinytip.co/tips/html-pdf-params/).
2. We can't implement a custom logic when the user clicks in links. The current behaviour is:
    a. Desktop: the iframe will navigate to the target URL
    b. Web: the **top level** url will change to the target URL (users can still CMD+Click to open in a new tab)

The pros are:
1. The PDF preview is native, so it's familiar to the user and is very smooth

hammer:
> * Can't interact with the PDF content (is an `<embed .... />`), we can only configure [how the viewer looks](https://tinytip.co/tips/html-pdf-params/).

Thanks for exploring @cguedes! I have ambitions for Ref Studio users to be able to link document content to reference content, so I'd prefer to follow the current path if possible.

cguedes:
@hammer @gjreda I was able and keep using **react-pdf** and fix the issue by adding a [polyfill](https://polyfill.io/v3/polyfill.min.js?features=Promise.allSettled%2CArray.prototype.at) and adopt the legacy PDF runner as they recommend in the docs (for when we need to support older browsers).

I've tested the release in the apple intel chip and is working 🎉  

Now I will wrap-up the PR by adding the other os variants and in order to prepare the work for #540

gjreda:
@cguedes For the sake of wrapping up this PR, do you want to add the other OS variants in a follow up PR?

I also think we should make some minor changes here so that this workflow is only triggered on a given `tag` or release flag. I was triggering it on all pushes simply for the sake of debugging.

cguedes:
I think I should add the two Apple variants to this PR (almost there actually).
I'm also changing the GH Action to be manually triggered (in the cut-release.yml).

image

cguedes:
I was able to release a version ([23.2.0-alpha.1](https://github.com/refstudio/refstudio/releases/tag/untagged-12d375f62bf0e4ead449)) with assets for Mac Intel chip and mac Silicon with success.

Will add more information about the release process in #540 

| Mac Intel chip | Mac silicon chip |
|--------|--------|
| image | image | 
| image | image
 |

.github/workflows/cut-release.yml
```diff
@@ -1,46 +1,133 @@
...
+      - uses: actions/setup-node@v3
 
-  setup-python:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-      - uses: ./.github/actions/setup-python
+      - name: get version
+        run: echo "PACKAGE_VERSION=$(node -p "require('./package.json').version")" >> $GITHUB_ENV
```
sehyod, line 22: You could keep these jobs so they run in parallel with `create-release` and so that all jobs `build-tauri` jobs can reuse the cache (don't forget to add `needs: [create-release, setup-js, setup-python, setup-rust]`

cguedes:
When I downloaded the .dmg I've got the error: "...is damaged and can’t be opened. You should move it to the Trash"

Found a [solution online that worked](https://osxdaily.com/2019/02/13/fix-app-damaged-cant-be-opened-trash-error-mac/) using `xattr -cr /Applications/refstudio.app`

sehyod (approved):

gjreda:
LGTM!

cguedes (approved):

PR #514, opened 2023-09-07
feat: update command palette design

### LINKED ISSUE #439: Design for command palette

sehyod:
This implements the new design for the command palette.
This hides actions depending on the context. The test file has been updated to reflect this change and make it easier to grow the component by adding new actions

Screenshot 2023-09-07 at 12 07 46

Closes #439


cguedes:
@sehyod I think we should add `h-6 w-6 p-[0.25rem]` to this rule to make sure the icons and text are aligned. Please confirm if this is the correct way to implement this.

```css
.command-palette .info_wrapper > svg {
   @apply h-6 w-6 fill-btn-ico-modal-item-default p-[0.25rem];
}
```

| Before | After |
|--------|--------|
| image | image |

sehyod:
Yes, this works. I did not update it because I thought we would use custom icons soon, but the design team confrimed we should use the icons from react-icons, so I will update the CSS rules

cguedes (approved):

Issue #301, opened 2023-07-13
Citation key scope, editing, and repair


hammer:
We had an interesting conversation a few weeks ago about what should happen to document text when a citation key is edited. It would be nice if all documents with citations using that citation key could update themselves. When thinking about a solution to this question, it became clear that we need to figure out if references are scoped to a document, project, or user. This issue is a long-term consideration but will have an impact on the backend so I'm filing now to ensure it's considered some time in the future.

sehyod:
Should we close this issue?
References are currently scoped by project, and we currently use the reference id in documents so the citation keys are automatically updated when editing them.

Exported files (markdown and bib) are not automatically updated and you need to be export them again

Issue #339, opened 2023-08-01
Incorporating DOI field in the UI and making it editable


shauryr:
Re: https://github.com/refstudio/refstudio/pull/338#issuecomment-1660508106

We need the ability to display the `doi` for each paper. Given that GROBID extraction could fail or may contain inaccuracies for some PDFs, granting users the capability to manually edit the `doi` will be a beneficial alternative.

PR #521, opened 2023-09-07
339 Incorporating DOI Field

jvoytek:
Implements [#339](https://github.com/refstudio/refstudio/issues/339)

- adds DOI to reference details view
- makes DOI and other string-type variables in that view editable (except [citationKey](https://github.com/refstudio/refstudio/issues/301))


src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,261 @@
...
+  let authorsString = '';
+  authorsArray.forEach((a, index) => {
+    authorsString += (index > 0 ? ', ' : '') + a.fullName;
+  });
+  return authorsString;
+};
+
+const getContentString = (content: string | Author[] | undefined, id: string): string => {
+  if (typeof content == 'undefined') {
+    return '';
```
sehyod, line 29: You can define this function like this `const getContentString = (id: string, content?: string | Author[])` so you don't have to write `| undefined`

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,261 @@
...
+};
+
+const DataTextInput = ({
+  content,
+  id,
+  onChangeHandler,
+}: {
+  content: string;
+  id: string;
+  onChangeHandler: CallableFunction;
```
sehyod, line 46: Please use the explicit function type `(id: string, value: string) => void` instead of `CallableFunction`.
Also nit comment: the convention is usually to just call it `onChange` and in this case, maybe `onSave` or `onBlur` would be better since this is not called in the change handler of the input

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,261 @@
...
+);
+
+const UnEditableTableDataCell = ({ contentData, id }: { contentData: string | Author[]; id: string }) => {
+  const contentString = getContentString(contentData, id);
+
+  return (
+    <td className="w-auto p-5">
+      <span className="leading-[30px]">{contentString}</span>
+    </td>
+  );
```
sehyod, line 107: `EditableTableDataCell` and `UneditableTableDataCell` are very similar and only used in `TableDataCell`. I think you should not separate the code in two components and write everything in `TableDataCell` directly instead, like this:
```typescript
return <td className="w-auto p-5">
    {editing && editable && typeof contentData === 'string' ?
      <DataTextInput content={contentData} id={id} onSave={onChangeHandler} /> :
      <span className="leading-[30px]">{getContentString(contentData, id)}</span>
    }
  </td>;
```

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,261 @@
...
+const IconButtons = (header: boolean | undefined): ReactElement[] => {
+  if (header) {
+    return [
+      IconButton({ hiddenWhenEditing: true, icon: EditIcon(), title: 'Edit Reference' }),
+      IconButton({ hiddenWhenEditing: false, icon: CloseIcon(), title: 'Finished Editing Reference' }),
+    ];
+  } else {
+    return [];
+  }
+};
```
sehyod, line 138: Same comment as above, I think you should remove `IconButtons` and write this logic in `TableHeadCell`. Also please use the JSX syntax for components instead of regular function syntax.
```typescript
...
{content}
{header &&
  (editing ?
    <IconButton icon={<CloseIcon />} key="closeIcon" title="Finished Editing Reference" onClick={() => setEditing(false)} /> :
    <IconButton icon={<EditIcon />} key="editIcon" title="Edit Reference" onClick={() => setEditing(true)} />)
}
```

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,261 @@
...
+const IconButton = ({
+  hiddenWhenEditing,
+  title,
+  icon,
+}: {
+  hiddenWhenEditing: boolean;
+  title: string;
+  icon: ReactElement;
+}) => {
+  const { editing, setEditing } = useContext(EditingContext);
```
sehyod, line 149: `IconButton` should just be a UI component, so it shouldn't "know" if the reference is being edited or not. I would change its props to the following
```typescript
{
  title: string;
  icon: ReactElement;
  onClick: () => void;
}
```
and pass `setEditing` from `TableHeadCell`. With the snippet I shared in my previous comment, the `IconButton`s are already conditionally rendered so there is no need for `hiddenWhenEditing` props anymore. It's usually better, when possible, to not render a component rather than render it and hide it with CSS

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,261 @@
...
+          updatedReference.title = value;
+          break;
+        case 'doi':
+          updatedReference.doi = value;
+          break;
+      }
+      handleReferenceChanged(updatedReference);
+    },
+    [updatedReference, handleReferenceChanged],
+  );
```
sehyod, line 197: Could you create `updatedReference` inside `referenceChanged`: no need to copy the reference on every render, you can do it only when we actually need to update it.
Also nit: the convention for such a function name is usually something like `handleReferenceChange`

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,261 @@
...
+          'text-card-txt-primary',
+        )}
+      >
+        <thead>
+          <tr>
+            <TableHeadCell colSpan={2} content="References" header={true} />
+          </tr>
+        </thead>
+        <tbody className="divide-y">
+          {Object.entries(referenceDetailsCardFormat).map((row) => {
```
sehyod, line 236: nit: You can write `{Object.entries(referenceDetailsCardFormat).map(([key, rowData]) => {`

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,261 @@
...
+            const key = row[0];
+            const rowData = row[1];
+            const contentData = reference[key as keyof ReferenceItem];
+
+            if (!contentData) {
+              return;
+            }
+
+            return (
+              <tr key={'row' + key + new Date().getTime().toString()}>
```
sehyod, line 246: Use `key={key}` directly or `key={'row' + key}` if you want the prefix (but we don't need it). You should not use the time because we want React to understand which row is being updated during re-renders. With the timestamp in the key, React would unmount and remount every row

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,261 @@
...
+        </thead>
+        <tbody className="divide-y">
+          {Object.entries(referenceDetailsCardFormat).map((row) => {
+            const key = row[0];
+            const rowData = row[1];
+            const contentData = reference[key as keyof ReferenceItem];
+
+            if (!contentData) {
+              return;
+            }
```
sehyod, line 243: You should remove this: when `contentData` is undefined, we still want to display the row, with an empty field, and this should be handled by `TableDataCell`

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,261 @@
...
+  );
+};
+
+const TableDataCell = ({
+  contentData,
+  id,
+  onChangeHandler,
+  editable,
+}: {
+  contentData: string | Author[];
```
sehyod, line 72: The type should be `contentData?: string | Author[];` here

sehyod (changes requested):
Thank you for your PR!

We may want to add unit tests, to check that the API is called when we expect for example.
Also the list of authors is currently not editable, I understand this is trickier, maybe we can create a follow-up issue for that?

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,261 @@
...
+  },
+});
+
+const returnFormatedAuthorsString = (authorsArray: Author[]): string => {
+  let authorsString = '';
+  authorsArray.forEach((a, index) => {
+    authorsString += (index > 0 ? ', ' : '') + a.fullName;
+  });
+  return authorsString;
+};
```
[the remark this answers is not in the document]
jvoytek: Thanks, I didn't know this existed!

src/features/references/editor/ReferenceDetailsCard.tsx
jvoytek: Now passing an empty string in case of no content.

sehyod: Perfect, that's even better! 👍

src/features/references/editor/ReferenceDetailsCard.tsx
jvoytek: ...
yeah turns out this was the thing causing the re-render to steal user focus after update. React is in fact smart enough not to re-render every row...now I know.

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,267 @@
...
+      IconButton({
+        hiddenWhenEditing: false,
+        icon: CloseIcon(),
+        title: 'Finished Editing Reference',
+        editing,
+        setEditing,
+      }),
+    ];
+  }
+};
```
sehyod, line 140: You should move this logic in `TableHeadCell`, like this:
```suggestion
    {header && setEditing && (
      editing ?
        <IconButton editing={false} icon={<CloseIcon />} setEditing={setEditing} title="Finished Editing Reference" /> :
        <IconButton editing={true} icon={<EditIcon />} setEditing={setEditing} title="Edit Reference" />
    )}
  </th>
);
```

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,267 @@
...
+  title,
+  icon,
+  editing,
+  setEditing,
+}: {
+  hiddenWhenEditing: boolean;
+  title: string;
+  icon: ReactElement;
+  editing: boolean;
+  setEditing: CallableFunction;
```
sehyod, line 153: `IconButton` should just be a UI component, so it shouldn't "know" if the reference is being edited or not. You should change its props to the following

```typescript
{
  title: string;
  icon: ReactElement;
  onClick: () => void;
}
```
With the snippet I shared in my previous comment, the `IconButtons` are already conditionally rendered so there is no need for `hiddenWhenEditing` props anymore

sehyod (changes requested):
Thank you for updating the PR and for the tests!
I still have some comments regarding the React code, I think you missed some of them because github hides them when there are too many. I have either bumped them or added new comments for the one I would like you to address before merging

src/atoms/__tests__/referencesState.update.test.ts
```diff
@@ -1,11 +1,20 @@
...
 
-import { updateProjectReference } from '../../api/referencesAPI';
+import * as referencesAPI from '../../api/referencesAPI';
 import { REFERENCES } from '../../features/references/__tests__/test-fixtures';
 import { ReferenceItem } from '../../types/ReferenceItem';
 import { getDerivedReferenceAtom, setReferencesAtom, updateReferenceAtom } from '../referencesState';
 
-vi.mock('../../api/referencesAPI');
+vi.mock('../../api/referencesAPI', async () => {
+  const actual: object = await vi.importActual('../../api/referencesAPI');
```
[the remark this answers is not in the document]
jvoytek: Mocking the whole module also mocks UPDATABLE_FIELDS which is now also used by ReferencesState for testable functions. I used importActual to mock only the functions that need to be mocked to avoid overwriting UPDATABLE_FIELDS. When using importActual, checking the mocked function directly results in this error: "[Function updateProjectReference] is not a spy or a call to a spy!" I don't know why. Updated to use importOriginal as first argument rather than importActual direct.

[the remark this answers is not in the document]
sehyod: Ok! That makes sense.

About the error you got, you can write
```typescript
{
    ...actual,
    updateProjectReference: vi.fn(),
}
```
so that the function is a mocked function instead of a regular function, and so you can check it in tests. But I don't mind the spy version

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,267 @@
...
+  id: string;
+  onChangeHandler: CallableFunction;
+}) => {
+  const [value, setValue] = useState(content);
+
+  return (
+    <input
+      className="w-full border bg-slate-50 px-2 py-0.5"
+      name={id}
+      title={id} // required for unit testing to find this element
```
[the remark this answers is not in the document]
jvoytek: I don't think that's quite right either. getByTestId should be a last resort but findByRole was not working as expected without a matching title attribute (for some reason??). In order to make these more resilient and more "like a user" would use the app I'm now using findByLabelText and globally unique IDs to the inputs so "for" attributes in the labels work. (Globally unique so that even if we end up putting more than one editable reference details card on the screen at a time the labels will still be associated with the correct inputs). This also improves accessibility.

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,267 @@
...
+  setEditing,
+}: {
+  content: string;
+  colSpan?: number;
+  header?: boolean;
+  id: string;
+  editing?: boolean;
+  setEditing?: CallableFunction;
+}) => {
+  const colSpanString = { ['colSpan']: colSpan };
```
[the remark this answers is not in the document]
jvoytek: It was my first attempt at a conditional attribute. I didn't want to clutter the DOM with colspan="1", but apparently React is smart enough not to render an attribute at all if the value is undefined. Fixed colSpan={colSpan}

jvoytek:
@sehyod Comments (including hidden ones) addressed.

jvoytek:
Should I have set these comments to "resolved" or waited for you to do that @sehyod? Still getting the hang of code review in github.

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -0,0 +1,267 @@
...
+      }
+      handleReferenceChange(updatedReference);
+    },
+    [reference, handleReferenceChange],
+  );
+
+  const referenceDetailsCardFormat = {
+    citationKey: {
+      title: 'Citation Key',
+      editable: false,
```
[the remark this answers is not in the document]
sehyod: We store the reference id in documents, so editing the citation key is already automatically reflected in them. I think we should close #301
If we decide to make the field non-editable, you should remove it from `UPDATABLE_FIELDS`.

sehyod:
> Should I have set these comments to "resolved" or waited for you to do that @sehyod? Still getting the hang of code review in github.

I usually "resolve" comments I address directly and let those I reply to unresolved so the person can more easily see my replies.

I am not approving yet because I think we need to decide what to do with the citation key field first: either making it editable or removing it from `UPDATABLE_FIELDS`. I personally think it should be editable.
Apart from that everything looks good, thank you for addressing my comments!

src/features/references/editor/ReferenceDetailsCard.tsx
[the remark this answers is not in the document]
cguedes: I agree that we already have a good support for editing the reference citation key. So I think we should keep the field editable. 

@sehyod do we use the reference ID to identify the reference item inside TipTap?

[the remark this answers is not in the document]
sehyod: Yes, when I say documents, I mean TipTap documents. I have added a comment to the issue #301. I think we can close it now and keep the citation key editable

hammer:
Okay if y'all feel strongly we can edit cite keys without issue I'm fine to keep them editable

sehyod (approved):

PR #522, opened 2023-09-08
feat: Add missing api response types

### LINKED ISSUE #510: Add missing API response types
After #414 and #457 we still have some API responses missing in the generated `api-paths.ts`. 

The `POST /api/projects` is an example this.

https://github.com/refstudio/refstudio/blob/28e9047ba389b1030c4cd6dbfa8dd7bc3f7a6a4c/python/sidecar/projects/router.py#L27-L42

https://github.com/refstudio/refstudio/blob/28e9047ba389b1030c4cd6dbfa8dd7bc3f7a6a4c/src/api/api-paths.ts#L602-L622

gjreda:
fixes #510 

@cguedes the following endpoints are still showing up as `unknown`:
- `read_file_api_fs__project_id___filepath__get` -- I suspect this has something to do with this endpoint returning fastapi's `FileResponse` rather than our own definition. I tried wrapping `FileResponse`, but no luck.
- `head_file_api_fs__project_id___filepath__head` -- doesn't return anything
- `shutdown_api_meta_shutdown_post` -- doesn't return anything

### DIFF
--- a/python/sidecar/projects/schemas.py
+++ b/python/sidecar/projects/schemas.py
@@ -1,25 +1,27 @@
-from __future__ import annotations
-
-from sidecar.typing import RefStudioModel, ResponseStatus
-
-
-class FileEntryBase(RefStudioModel):
-    name: str
-    path: str
+from sidecar.filesystem.schemas import FileEntry, FolderEntry
+from sidecar.typing import RefStudioModel
 
 
-class FileEntry(FileEntryBase):
-    file_extension: str
-
-
-class FolderEntry(FileEntryBase):
-    children: list[FileEntry | FolderEntry] = []
-
-
-class ProjectDetailsResponse(RefStudioModel):
+class ProjectBase(RefStudioModel):
     id: str
     name: str
+
+
+class Project(ProjectBase):
     path: str
+
+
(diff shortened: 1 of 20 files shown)


python/sidecar/projects/schemas.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import annotations
```
gjreda, line 1: @cguedes I slightly changed a few of the `projects` related return types, so let me know if you have any issues with this. When we were first writing this backend, we were just returning `str` and `dict` all over, so did my best to make proper types.

gjreda: I also pushed with `--no-verify` as the `projects` type changes resulted in some small issues on the TS side, but I can give fixing those a shot once we're aligned on the changes here.

cguedes (approved):
This is ready to go 🚀 

I will see what we can do for filesystem API output files in order to adopt the new api design Dan started.

⚠️ This PR has a breaking change for the internal storage format for projectes. So everyone needs to be aware that they should `yarn reset:storage` after pulling main.

To help, I updated the reset script to backup the content of the folder `user1` so that we can copy paste some files we might have there.

Issue #59, opened 2023-05-28
Local AI interactions


hammer:
Another topic to lift from Notion to GitHub...

https://github.com/ggerganov/llama.cpp and https://github.com/ggerganov/whisper.cpp seem to have the most active development. Both are built on top of the underlying https://github.com/ggerganov/ggml library. https://github.com/abetlen/llama-cpp-python is a Python client library for `llama.cpp`.

In Rust, there's https://github.com/rustformers/llm, built on the [ggml](https://github.com/rustformers/llm/tree/main/crates/ggml) crate that wraps the ggml C library. https://github.com/rustformers/llm/discussions/71 is an interesting discussion of the history of the Rust libraries.

https://github.com/jankais3r/LLaMA_MPS is an alternative implementation in Python with different performance characteristics.

https://github.com/mlc-ai/mlc-llm seems to focus more on mobile devices and https://github.com/mlc-ai/web-llm uses WebGPU to run in the browser. The [Apache TVM](https://tvm.apache.org/) community seems to be leading this work. [OctoML](https://octoml.ai/) is commercializing TVM.

Local implementations of chat interfaces include
- https://github.com/imartinez/privateGPT
- https://github.com/nomic-ai/gpt4all
- [LM Studio](https://lmstudio.ai): closed source

hammer:
Looks like `llama.cpp` will use the gpu soon https://github.com/ggerganov/ggml/pull/108

hammer:
Another local chat codebase: https://github.com/Birch-san/llama-play

hammer:
This looks like some Go code that wraps local LLMs to give them a common interface on the "server" side: https://github.com/go-skynet/LocalAI

hammer:
Python wrapper for ggml: https://github.com/marella/ctransformers

hammer:
Guide from W&B: https://wandb.ai/capecape/LLMs/reports/A-Guide-on-Running-LLMs-Locally--Vmlldzo0Njg5NzMx

hammer:
https://github.com/jmorganca/ollama looks handy for macOS

hammer:
https://github.com/Bin-Huang/chatbox is a Tauri app that allows local models. We might get some inspiration from their implementation

hammer:
[LM Studio](https://lmstudio.ai/), mentioned above, has a repo with examples of using its service interface: https://github.com/lmstudio-ai/examples.

Issue #58, opened 2023-05-28
Python client for AI interactions


hammer:
We've talked a bit about choosing a Python client that can be used across LLM APIs, as well as potential abstractions for working with prompt templates and prompt chains. It's probably time to take those conversations from Notion and Slack to GitHub!

A big list of prompt engineering tools is at https://www.promptingguide.ai/tools. I don't think we need to explore all of them of course. Many tools belong to more than one of these categories, but I think it's useful to break them down a bit.

### Client libraries
These are libraries that provide a common Python interface to multiple LLM backends.

- LangChain's [Models](https://python.langchain.com/en/latest/modules/models.html)
- LlamaIndex's [LLM](https://gpt-index.readthedocs.io/en/latest/core_modules/model_modules/llms/root.html)
- https://github.com/r2d4/openlm
- https://github.com/danielgross/python-llm
- https://github.com/kagisearch/pyllms

### Prompt templates
These are libraries that make working with complex prompts easier, usually using some template library like Jinja or Handlebars.

- LangChain's [Prompts](https://python.langchain.com/en/latest/modules/prompts.html)
- https://github.com/microsoft/guidance
- https://github.com/HazyResearch/manifest

### Call trees
These are libraries that make it easier organize and execute a collection of calls to LLM and potentially other APIs (often vector databases) necessary to solve a high-level task. It would probably be better to break this group up to separate indexing and memory APIs from call tree APIs...

- LangChain's [Chains](https://python.langchain.com/en/latest/modules/chains.html)
- https://github.com/srush/MiniChain
- https://github.com/jerryjliu/llama_index
- https://github.com/eth-sri/lmql
- https://github.com/stanfordnlp/dsp

hammer:
Nice work not yet in a separate library https://twitter.com/gneubig/status/1663544322233606148

hammer:
https://github.com/deepset-ai/prompthub for prompt management

hammer:
https://github.com/hegelai/prompttools

hammer:
https://github.com/microsoft/Typechat for typed prompt building in TypeScript

hammer:
https://github.com/philschmid/easyllm another client library

hammer:
https://github.com/BerriAI/litellm another client library

hammer:
There's a fourth category I suppose for controlling output.

### Controlling output
- https://github.com/1rgs/jsonformer
- https://github.com/ShreyaR/guardrails
[...]

PR #525, opened 2023-09-08
feat: Add support for Ollama as a local LLM via litellm

cguedes:
related to #58 and #59

This PR adds support for using Ollama as a local LLM via `litellm`.

To use refstudio with [Ollama](https://ollama.ai/), you'll need to ensure that both the Ollama server and refstudio are running on your local machine at the same time.
1. Download and install Ollama
2. Launch the Ollama server locally: `ollama serve`
3. Verify the Ollama server is working: `ollama run llama2 "are you running?"`
4. Launch refstudio
5. In refstudio > settings, change your `chat_model` setting to `llama2`
_____________
This is an experimental PR trying to integrate [LiteLLM](https://github.com/BerriAI/litellm) with RefStudio with a local LLM using [Ollama](https://github.com/jmorganca/ollama).

**Notes:**
* ~🤔 I'm not sure if the async/await should be used here. I needed to use that (and the unwrap for the local LLM. And then implemented the same call (`litellm.acompletion`) for the OpenAI~
* ~💥 I haven't fixed unit test. They need to be adjusted to use the new async methods~
* ⚠️ We still need to find a way to configure the open AI api key. I think that the litellm would fallback to the env var, something we don't want
* ℹ️ I've made simple to toggle between `gpt-3.5-turbo` and `llama2` in the settings. I think we still want to keep the input field, and we don't have a dropdown+input form element (the one that allow to select an option but also write any text we want).
* ~💥 The desktop can't start with LiteLLM! I've `yarn python` and then running `yarn tauri:dev:debug` would show the following~

To run this example, first install Ollama (with llama2) and start it:

```bash
ollama serve
ollama run llama2 'are you running?'
```

To test inside Refstudio (in the rewrite) you just need to configure the correct model in the settings before each rewrite. You can use `llama2` (local Ollama) or `gpt-3.5-turbo` (for OpenAI - you need the API Key correctly configured).

image

image


sehyod (commented):
You should update the comment of the `Rewriter` class (lines 164 to 177 in rewrite.py)

sehyod (commented):
The frontend part looks great!
The backend part does not work when running the desktop app (the server cannot start so the app doesn't start). And as discussed offline, I don't understand what API key we are using, but even with an invalid key in my settings, the requests are successful

gjreda:
Chatted with @cguedes offline and I'll plan to pick up this PR on Monday and work through the remaining backend issues.

ishaan-jaff:
@cguedes i'm the maintainer of litellm sorry for the errors and poor experience you encountered:
Looks like there are two main action items to improve litellm here (please let me know if there are more)

- yarn python fails with litellm looks like `get_encoding fails` and then leads to a bunch of errors 
- Don't fallback to .env for checking keys @cguedes how would you pass keys to litellm instead ?

gjreda:
@cguedes @sehyod IMO all that's left here is:
1. Rename the `openai` related backend settings to be more generic. I'm thinking we'll have:
- `provider` - The LLM service/company to use (e.g. `openai`, `anthropic`, `ollama`)
- `api_key` - The API key for the specified provider (`None` for ollama)
- `model` - No need to call this `chat_model` since it is used by all AI interactions at this point
2. Update the frontend settings accordingly.

Also not sure why the prettier check is failing since it runs fine locally and on the pre-commit.

.github/workflows/test-sidecar-binary.yml
```diff
@@ -0,0 +1,32 @@
+name: Build and test sidecar binary
```
gjreda, line 1: Feels like we should always be testing this when python files are changed

cguedes:
@gjreda the client side code is ready. This is the updated settings panel.

image

gjreda (approved):

src/features/ai/settings/AiSettingsPane.tsx
```diff
@@ -52,10 +66,42 @@ export function OpenAiSettingsPane({ config }: SettingsPaneProps) {
...
+        />
+        <InfoText>
+          <strong>Use: </strong>
+          <InfoTextOptions
+            options={[
+              { value: 'gpt-3.5-turbo', label: 'OpenAI' },
+              { value: 'llama2', label: 'for Ollama' },
+            ]}
+            onClick={(value) => setPaneSettings({ ...paneSettings, model: value })}
+          />
```
sehyod, line 102: My understanding of the backend code is that the model is overridden, to use 'llama2' when the provider is ollama. Maybe we should reflect that in the frontend by disabling the field and displaying 'llama' when selecting ollama provider?

sehyod (commented):
I have tried the branch locally and I have 2 comments:
 - I don't think we are properly using the API key from the settings: even with an invalid key (or without any key at all) I am still able to rewrite text
 - I haven't been able to use ollama provider: I have installed everything, checked the server was working with `ollama run llama2 "are you running?"` and I get an error when trying to rewrite my selection (`RetryError[<Future at 0x10f557210 state=finished raised TypeError>]`)

cguedes:
> I have tried the branch locally and I have 2 comments:
> 
> * I don't think we are properly using the API key from the settings: even with an invalid key (or without any key at all) I am still able to rewrite text
> * I haven't been able to use ollama provider: I have installed everything, checked the server was working with `ollama run llama2 "are you running?"` and I get an error when trying to rewrite my selection (`RetryError[<Future at 0x10f557210 state=finished raised TypeError>]`)

@sehyod thanks for the feedback.  

1. The library fallback to an env var. Will check that behaviour again
2. That is fixed that in my latest commit

sehyod:
> 1. The library fallback to an env var. Will check that behaviour again
> 2. That is fixed that in my latest commit

1. I don't have any .env file so it doesn't explain how it works, and I think it should not fallback to an env var if the user provides a key through the app, even if the key is not valid
2. Thanks for the fix! I confirm it is working now 🎉

src/features/ai/settings/AiSettingsPane.tsx
cguedes: I will make the llama2 default when we select Ollama and same for OpenAI. They can use other values like gpt4 or other ollama models.

python/sidecar/ai/chat.py
```diff
@@ -22,12 +23,15 @@ def ask_question(
...
-        openai.api_key = user_settings.openai_api_key
-        model = user_settings.openai_chat_model
-    else:
-        openai.api_key = os.environ.get("OPENAI_API_KEY")
+    if user_settings is None:
+        # this is for local dev environment
+        openai.api_key = os.environ.get("API_KEY")
         model = "gpt-3.5-turbo"
+    elif user_settings.model_provider == ModelProvider.OPENAI:
+        openai.api_key = user_settings.api_key
```
sehyod, line 31: I think this should be `os.environ["OPENAI_API_KEY"] = user_settings.api_key`: openai.api_key doesn't seem to be used, which is why I was getting a response even when my key was invalid

sehyod: After more investigation, it appears that when `openai.api_key` is invalid, liteLLM falls back to the `OPENAI_API_KEY` env var. I have no clue why my env var is correct (I don't have an .env file) but this is what was happening.
This behavior is a bit confusing, I think we should not fallback to anything if there is an api key set in the app. Carlos told me that the plan was to send the api_key in the request, instead of using `openai.api_key`, maybe this would prevent the fallback. In any case, this is a very edge case because the user has to have an `OPENAI_API_KEY` env var for the fallback to happen, so I don't think it should block this PR, but we have to keep it in mind

sehyod (approved):

PR #532, opened 2023-09-11
feat: Explorer panel design for multiple projects

cguedes:
This PR updated the UI to the latest design with menu triggers for the project and files.

image

image


src/components/FileNameInput.tsx
```diff
@@ -8,9 +8,10 @@ interface FileNameInputProps {
   onCancel: () => void;
   onSubmit: (newValue: string) => void;
   isNameValid: (name: string) => boolean;
+  className?: string;
 }
 
-export function FileNameInput({ fileName, isNameValid, onCancel, onSubmit }: FileNameInputProps) {
+export function FileNameInput({ className, fileName, isNameValid, onCancel, onSubmit }: FileNameInputProps) {
```
sehyod, line 14: I would rather not add a `className` prop to this component. You can update the padding here directly (and I have actually done it in #531 so we will have conflicts)

src/application/sidebar/FileExplorer.tsx
```diff
@@ -51,29 +73,9 @@ export function FolderNode({ folder }: FolderNodeProps) {
...
-              <FolderNode folder={child} key={child.path} />
-            ) : (
-              <FileNode existingFileNames={childrenEntries.map((file) => file.name)} file={child} key={child.path} />
-            ),
-          )}
-        </div>
-      )}
+    <div className="flex flex-col items-start gap-1 self-stretch">
+      <ProjectExplorerChevronNode collapsed={collapsed} text={folder.name} onClick={() => setCollapsed((v) => !v)} />
+      <ProjectExplorerEntries collapsed={collapsed} projectFileEntries={childrenEntries} />
```
sehyod, line 78: Could you do `{!collapsed && <ProjectExplorerEntries...}` instead? And remove `collapsed` from `ProjectExplorerEntries` props

src/application/sidebar/FileExplorer.tsx
```diff
@@ -8,41 +8,63 @@ import {
...
-import { useFileExplorerContextMenu } from './fileExplorerContextMenu/useFileExplorerContextMenu';
-import { ArrowRightIcon } from './icons';
+import { FILE_EXPLORER_PROJECT_MENU_ID } from './fileExplorerContextMenu/FileExplorerProjectContextMenu';
+import {
+  useFileExplorerContextMenu,
+  useFileExplorerOptionalContextMenu,
+} from './fileExplorerContextMenu/useFileExplorerContextMenu';
+import { ArrowDownIcon, HorizontalDotsIcon } from './icons';
+
+export function FileExplorer({ projectName }: { projectName: string }) {
```
sehyod, line 26: nit: maybe this should be renamed to something like `ProjectNode` now?

src/application/sidebar/FileExplorer.tsx
```diff
@@ -8,41 +8,63 @@ import {
...
+  const [collapsed, setCollapsed] = useState(false);
```
sehyod, line 27: This is stored in an atom for projects folders, so that when you close and re-open the explorer, everything kept its state. I think it would be great to also save that in an atom for the projects.
I think we should create an atom `openProjectState` with some useful values like `name`, `id` and `isCollapsed`. When we update the app so support multiple projects, we can then update the atom to be a list of `ProjectState`s

sehyod (changes requested):

src/components/FileNameInput.tsx
cguedes: Ok. Will revert my change and wait for your PR get approved.

src/application/sidebar/FileExplorer.tsx
cguedes: I see your idea, but I think we should implement that later once we really support having multiple projects open at the same time. For now we are only making sure we have the design ready.

sehyod (approved):

hammer:
As part of this change can we create a top-level folder called "documents", put it above exports and uploads, and put the documents under there?

hammer:
Or if that breaks the data model, we should at least have the documents above the exports and uploads, I think.

cguedes:
That change of having a top-level `documents` folder is a bit complex in terms of data model and UI logic to make it expanded (when opening a project, creating a new file, ...).

But showing the files before folders (i.e. above exports and uploads) should be very simple. We will do that in a follow-up PR.

PR #539, opened 2023-09-12
feat: update references table styles

### LINKED ISSUE #378: Design for References Table Editor

sehyod:
This updates the references table to the latest design. 
This also creates all CSS variables and updates the tailwind color names to use these new variables

Screenshot 2023-09-12 at 13 05 56

The empty view uses the empty behaviour of ag grid:
image

Closes #378


src/features/references/editor/__tests__/ReferencesTableView.test.tsx
```diff
@@ -117,24 +118,21 @@ describe('ReferencesTableView component', () => {
...
     await user.click(removeBtn);
 
+    expect(removeBtn).toBeDisabled();
     expect(removeBtn).toHaveAttribute('aria-disabled', 'true');
     expect(vi.mocked(emitEvent)).not.toHaveBeenCalled();
   });
 
-  // Note: This test can't run because there is a weird issue when clicking remove ("TypeError: Cannot read properties of undefined (reading 'contains')")
-  // https://github.com/ag-grid/ag-grid/issues/6179
-  it.skip(`should emit ${'refstudio://references/remove' as RefStudioEventName} with SOME selections`, async () => {
```
sehyod, line 131: I don't know what was the issue and what fixed it, but I tried to unskip the test and it passed! 🎉

hammer:
cc @jvoytek as I think she was going to take a look at doing this one as well

src/features/references/editor/ReferencesTableView.tsx
```diff
@@ -142,64 +134,75 @@ export function ReferencesTableView({ defaultFilter = '' }: { defaultFilter?: st
...
-          <TopActionIcon
-            action="Reset Cols"
-            icon={VscTable}
-            onClick={() => resetColumnsState(() => gridRef.current!.columnApi.resetColumnState())}
+          <Button
+            Action={<ExportIcon />}
+            size="M"
+            text="Export"
+            type="secondary"
+            onClick={() => emitEvent('refstudio://menu/references/export')}
```
cguedes, line 171: I think that this should behave like the Remove, exporting the selected references.

cguedes (changes requested):

src/components/Button.tsx
jvoytek, line 1: Consider removing `disabled={disabled}` at line 55. Button will be "focusable" (allowing screen readers to find them and read information like why is this disabled if available) but functionality will still be disabled onClick (line 58) and appear "grayed out" (lines 47 && 51). aria-disabled (only) is preferred for (accessibility)[https://css-tricks.com/making-disabled-buttons-more-inclusive/#:~:text=The%20difference%20between%20disabled%20and%20aria%2Ddisabled&text=The%20only%20overlap%20between%20the,disabled%20is%20all%20about%20semantics.]

src/features/references/editor/ReferencesTableView.css
jvoytek, line 1: Just a note for future development, using @apply here will require additional style sheets when implementing dark mode. I don't believe AG Grid can be themed another way so this is how it will have to be, but in general preferring TailwindCSS utility classes in-line will allow for (easier dark mode theming in the future) [https://tailwindcss.com/docs/dark-mode]

src/features/references/editor/ReferencesTableView.tsx
jvoytek: I agree.

src/features/references/editor/ReferencesTableView.tsx
```diff
@@ -142,64 +134,75 @@ export function ReferencesTableView({ defaultFilter = '' }: { defaultFilter?: st
...
           />
         </div>
+        <div className="h-8 w-px bg-content-area-border" />
+        <Button
```
jvoytek, line 175: It seems odd that this is such a large button with such a small bit of functionality that only does anything if you've messed with the column widths. Not necessarily something to fix right now but worth thinking about?

src/features/references/editor/__tests__/ReferencesTableView.test.tsx
```diff
@@ -117,24 +118,21 @@ describe('ReferencesTableView component', () => {
   it(`should not emit ${'refstudio://references/remove' as RefStudioEventName} with NO selections`, async () => {
     const { user } = setupWithJotaiProvider(<ReferencesTableView />, store);
 
-    const actionsMenu = screen.getByTestId('actions-menu');
-    await user.hover(actionsMenu);
-    const removeBtn = within(actionsMenu).getByText('Remove');
+    const removeBtn = screen.getByText('Remove');
     await user.click(removeBtn);
 
+    expect(removeBtn).toBeDisabled();
```
jvoytek, line 124: Assuming you remove `disabled={disabled}` in the button component this would also be removed.

jvoytek:
> cc @jvoytek as I think she was going to take a look at doing this one as well

Reviewed and made some comments.

src/components/Button.tsx
sehyod: TIL
Thank you!

src/features/references/editor/ReferencesTableView.tsx
cguedes: This is a good point @jvoytek I only added this initially because I was messing around a lot with columns. I agree that it can be smaller (ex: under a "3 vertical dots" menu popup).

sehyod: I'll share it with the design team

cguedes (approved):

Issue #441, opened 2023-08-29
Implement cheat sheet screen


sehyod:

cguedes:
The current design only has content for the first screen/tab/section.

![image](https://github.com/refstudio/refstudio/assets/174127/bfa4b4ee-bd5a-40eb-8879-5983313dbd4a)

We need to decide which tabs we will have and what is the content. 

This is the list of all keyboard shortcuts used by the application:

Ref Studio Keyboard Shortcuts
* Menu → File
    * `⌘ N` - New File (known shortcut)
    * `⌘ S` - Save File (known shortcut)
    * `⌘ W` - Close Editor (known shortcut)

* Menu → Edit
    * These are all standard know, native, shortcuts

* Menu → References
    * `⌘ R` - Open the references editor (our shortcut, normally used to refresh the app)

* Menu → View
    * Toggle Full Screen - Native shortcut. Don’t even know what that globe is!
    * `F11` - Notifications - Toggle the notification drawer (our shortcut)

* Menu → Window
    * `⌘ M` - Native shortcut to minimize the app

* Application
    * `⌘ K` - Open Command Palette (+/- known shortcut)
    * `⌘ P` - Open Files Palette (menu inside the command palette) (shortcut inspired by VSCode)
    * `⌘ 1` - Select / Toggle Visibility of Explorer Pane (our shortcut)
    * `⌘ 2` - Select / Toggle Visibility of References Pane (our shortcut)
    * `⌘ 9` - Select / Toggle Visibility of Rewriter (our shortcut)
    * `⌘ 0` - Select / Toggle Visibility of Chatbot (our shortcut)

* Editor
    * Standard text editor keyboard shortcuts (⌘ B, ⌘ I, ⌫, ⌘ Z, ⌘ ⇧ Z, ...)
    * Standard basic markdown keyboard [shortcuts](https://www.markdownguide.org/cheat-sheet/) (#, ##, ###, *, ...)
    * Text completion (our shortcuts)
        * `⌘ J` - Ask for text completion (when triggered again cycle through choices)
            * `↹`, `→`, `↵` - to accept suggestion
            * `⎋` - (Escape) to remove suggestion
    * References (our shortcuts)
        * `[` - Trigger a references citation interaction (show a popup to select a reference)
            * `@` - Inside a citation will trigger the popup to select a reference)
    * "Notion-like" shortcuts
        * `> {SPACE}` - Creates a collapsible section
[...]

Issue #346, opened 2023-08-14
Update UI to incorporate new YLD designs


hammer:

hammer:
FYI let's go with the navy blue color scheme for now

PR #546, opened 2023-09-14
feat: UI redesign review

cguedes:
This PR add multiple small improvements that the design team made over last week, and also the foundation UI elements and interaction for #441.

* Adjust design of homepage (initial run)
* Adjusted welcome screen when the user already has projects
* Configure font sizes (base 14px), (base editor 16px)
* Improve project side bar and Create/Select modals
* Improve references styles
* Add keyboard shortcuts modal

image

image

image

image

image

image

Related to #346


src/application/KeyboardShortcutsModal.tsx
```diff
@@ -0,0 +1,158 @@
...
+function Tabs({ selectedTab, onClick }: { selectedTab: ModalTab; onClick: (tab: ModalTab) => void }) {
+  return (
+    <div>
+      <div className="relative flex w-full pl-3 pt-1">
+        {MODAL_TABS.map((tab) => (
+          <div
+            className={cx('px-4 py-3', {
+              'font-bold': selectedTab.name === tab.name,
+              'z-[1] border border-b border-modal-border border-b-white': selectedTab.name === tab.name,
+              'cursor-pointer hover:font-bold': selectedTab.name !== tab.name,
```
sehyod, line 108: This is just a design preference, but I would keep the cursor-pointer even for the active tab

And a nit comment: in the design the tab borders are rounded and there is no top margin (the left and right border take the whole height and there is no top border. I don't mind the top border, but I think we should make everything rounded for consistency with the rest of the app, especially with the editor tabs.

sehyod (approved):

sehyod:
Thank you for this PR! I have small comments regarding the design but nothing blocking so I have approved it

src/application/KeyboardShortcutsModal.tsx
cguedes: Thanks for your feedback. We redesigned this screen so now we don't need tabs and round corners  🎉 

image

cguedes:
@sehyod this is ready for another review.

sehyod (approved):
Nice! Thank you for all these updates!

PR #547, opened 2023-09-14
feat: add endpoint for streaming chat

### LINKED ISSUE #541: Update chat to stream response from LLM

gjreda:
fixes #541

### DIFF
--- a/python/openapi.json
+++ b/python/openapi.json
@@ -933,7 +933,11 @@
               "application/json": {
                 "schema": {}
               },
-              "text/event-stream": {}
+              "text/event-stream": {
+                "schema": {
+                  "type": "string"
+                }
+              }
             },
             "description": "Stream chat reply."
           },
--- a/python/openapi.json
+++ b/python/openapi.json
@@ -42,6 +42,10 @@
             "default": 1,
             "type": "integer"
           },
+          "stream": {
+            "default": true,
+            "type": "boolean"
+          },
           "temperature": {
             "default": 0.7,
             "type": "number"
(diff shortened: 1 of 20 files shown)


gjreda:
@cguedes After taking a quick shot at the frontend for this, I realized this response is a new pattern relative to our other APIs (e.g. I don't think we can use the existing `typed-api.ts` for this). Figured it's probably a good idea to pair on the frontend with you or let you handle it.

python/openapi.json
```diff
@@ -903,6 +907,56 @@
...
+              }
+            }
+          },
+          "required": true
+        },
+        "responses": {
+          "200": {
+            "content": {
+              "application/json": {
+                "schema": {}
```
cguedes, line 938: This isn't correct. It's another "incompatibility" with the codegen and fast api internal types (here StreamingResponse).

python/sidecar/ai/schemas.py
```diff
@@ -61,6 +61,7 @@ class ChatRequest(RefStudioModel):
     text: str
     n_choices: int = 1
     temperature: float = 0.7
+    stream: bool = False
```
cguedes, line 64: Given that we will use a different endpoint, I think we don't need this in the request payload. 

With this approach maybe we can reuse the existing API endpoint `/api/ai/{project_id}/chat` and decide on the return type using that field.

Keeping this change and two endpoints, mean that we need to throw an error if someone call the `/api/ai/{project_id}/chat` endpoint with `stream: true`.

cguedes (changes requested):

python/sidecar/ai/schemas.py
gjreda: How about `ChatRequest` remains the same (i.e. remove `stream` property), but `ChatRequest` is used by both endpoints?

Internally within `/chat_stream` I'll pass in `stream: true` as these two endpoints share much of the same code, but ultimately need to fork a slightly different path.

cguedes: Exactly. That's a good solution IMO.

python/openapi.json
cguedes: To clarify, we can merge this. Calling this incorrect is a bit to much. I just wanted to note that we should try to have a way to type this response, but I understand how hard it can be ([swagger spec](https://github.com/OAI/OpenAPI-Specification/issues/1576)).

gjreda: I'm happy to add a one-off line in our schema generation code to handle this endpoint (and file upload) if you feel there's a correct way to handle it.

cguedes:
@gjreda I've added two commits with:

- change to the API output (media type and response format)
- update chat panel to use the new streaming API (some client tests are failing. will fix them in a follow-up commit)

More information in: https://medium.com/@hxu296/serving-openai-stream-with-fastapi-and-consuming-with-react-js-part-1-8d482eb89702

python/openapi.json
cguedes: Can we configure the stream method with the following:

```
@router.post(
    "/{project_id}/chat_stream",
    responses={
        200: {
            "content": { "text/event-stream": {} },
            "description": "Stream chat reply.",
        }
    },
)
```

This would make this information available here to be used in a follow-up issue to improve the "auto-api" in the client for text stream API calls. CC @danvk

cguedes:
@gjreda I've identified two issues:

- stream API for ollama
- error handing in the server (ex: try to use an invalid API key or model for open AI)

image

cguedes (approved):
This is ok for me.

PR #549, opened 2023-09-14
feat: Add endpoint for creating a new reference with or without PDF

### LINKED ISSUE #548: Add endpoint(s) for adding a new reference from S2 search results
This is at least two endpoints:
1. Takes a url to a pdf and downloads the pdf to a projects `uploads`. Possibly calls some ingest related functions (e.g. chunk reference).
2. Takes the metadata from S2 and updates the reference json storage accordingly

              Image preview for reference...
Screenshot 2023-09-13 at 11 39 55 AM
Screenshot 2023-09-13 at 11 41 47 AM

_Originally posted by @jvoytek in https://github.com/refstudio/refstudio/issues/138#issuecomment-1718140352_

gjreda:
fixes #548 

This PR adds a new endpoint to create a reference from some metadata (e.g. from S2 search results). If the reference metadata includes a PDF URL, the PDF will be fetched and added to the project uploads.

TODO:
- [x] Allow for adding a Reference _without_ a PDF
- [x] Default `source_filename` to `{title}.pdf`
- [x] Add citation key
- [x] ~Ability to parse Author in a more robust way?~ Rather than trying to parse on our own, we should probably just keep as `full_name` (which sometimes already happens with grobid) and allow the user to edit it in the UI if they'd like to change it.
- [x] Add publication date to S2 search results (would be good to pass this through with the metadata to the new endpoint)


python/sidecar/references/schemas.py
```diff
@@ -76,8 +89,30 @@ class ReferenceStatus(RefStudioModel):
...
+
+class IngestPdfUrlRequest(IngestRequest):
+    type: IngestRequestType = IngestRequestType.PDF
+    url: str
+    metadata: ReferenceCreate
+
+
+class IngestGenericRequest(IngestRequest):
+    type: IngestRequestType = IngestRequestType.GENERIC
+    metadata: ReferenceCreate
```
gjreda, line 115: Not quite sure how y'all would prefer to handle this on the frontend.

Since we are now allowing users to create references w/o PDFs, it might make sense for us to drop `IngestPdfUrlRequest` and instead just use `IngestGenericRequest` and add an optional `url` parameter. So two potential request types, rather than three.

Do you have thoughts @cguedes @jvoytek?

jvoytek: So long as available PDFs from S2 are still ingested 2 types is fine with me.

python/sidecar/references/service.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Returns
+    -------
+    ReferenceCreate
+        The metadata of the reference to create.
+    """
+    response = requests.get(url)
+
+    if not metadata.source_filename:
+        metadata.source_filename = f"{metadata.title}.pdf"
```
jvoytek, line 36: This can result in the filename being too long.

jvoytek:
Can we add to the to-do list for this pull to return the id of the newly created reference to facilitate removal on the S2 Search dialog (only right after addition ie user made a mistake).

python/sidecar/references/router.py
```diff
@@ -24,35 +30,39 @@ async def list_references(project_id: str) -> list[Reference]:
...
+        uploads_dir = get_project_uploads_path(user_id, project_id)
+        response = ingest.run_ingest(pdf_directory=uploads_dir)
+
+    elif request.type == IngestRequestType.PDF:
+        reference = create_reference(
+            project_id, metadata=request.metadata, url=request.url
+        )
+        response = IngestResponse(project_name=project_id, references=[reference])
+
     return response
```
gjreda, line 59: > Can we add to the to-do list for this pull to return the id of the newly created reference to facilitate removal on the S2 Search dialog (only right after addition ie user made a mistake).

@jvoytek This should be what already happens here. The endpoint returns an `IngestResponse` with a `references` key, which in this case would contain the newly created reference.

python/sidecar/references/schemas.py
cguedes: I don't mind 2 types. But would call it then `IngestMetadataRequest`, given that the PDF is now optional and generic is too abstract.

cguedes:
@gjreda can we return the `filepath` in the API (and even save in the metadata), and return undefined/None if the reference don't contain a PDF file?

Currently the create that value in the client and that won't work with the new optional PDF rule.

https://github.com/refstudio/refstudio/blob/79851279a9fcbd3bebbbc651eb2d3c1cfb56d9e5/src/api/referencesAPI.ts#L59-L63

CC @jvoytek

jvoytek:
I am so very pleased, references without a PDF work like magic this morning. A few things:

- [ ] If you haven't uploaded any PDF's and attempt to add using S2 first:
`  File "/Users/jvoytek/refstudio/python/sidecar/references/storage.py", line 49, in load
    with open(self.filepath, "r") as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/web-storage-url/user1/f82d3af8-48d5-478a-9542-e8603575f002/.storage/references.json'`

- [ ]   For S2 results with (what I assume is) a redirect for the openAccessPdf like this: http://www.thelancet.com/article/S1474442214703205/pdf pypdf chokes like this:
`File "/Users/jvoytek/Library/Caches/pypoetry/virtualenvs/refstudio-hQ57T0PE-py3.11/lib/python3.11/site-packages/pypdf/_utils.py", line 266, in read_previous_line raise PdfStreamError(STREAM_TRUNCATED_PREMATURELY)`

- [ ] For all references (even ones added in through upload) when you try to open the PDF:
`RuntimeError: File at path /tmp/web-storage-url/user1/9a8571db-a2a0-4a93-94a8-a6c29c6287c9/tmp/web-storage-url/user1/9a8571db-a2a0-4a93-94a8-a6c29c6287c9/uploads/Abbott-Neuron2008.pdf does not exist.`

jvoytek:
Any idea why this payload wouldn't save the published_date:
{
"type":"metadata",
"metadata":{
    "source_filename":"",
    "citation_key":"",
    "doi":"",
    "title":"Longitudinal changes in aperiodic and periodic activity in electrophysiological recordings in the first seven months of life",
    "abstract":"",
    "contents":"",
    "published_date":"2020-07-18",
    "authors":[
        {"full_name":"N. Schaworonkow"},
        {"full_name":"Bradley Voytek"}
    ]
}
}

gjreda:
> Any idea why this payload wouldn't save the published_date

I added it to search results, but forgot to add it in when creating the Reference.

> If you haven't uploaded any PDF's and attempt to add using S2 first

Should be fixed now. This feature breaks a core assumption we've had thus far -- that ingests are happening via the `uploads` ingestion call. With this feature, it's now possible that `.storage` might not exist since it's forking a different path than has been required before.

> For S2 results with (what I assume is) a redirect for the openAccessPdf

Need to look into this one a little more.

> For all references (even ones added in through upload) when you try to open the PDF

Still debugging this one. I'm able to view the PDF in some cases, but there are some edges where it still doesn't work. There might be some inconsistencies in how some frontend views send the filepath back.

python/sidecar/shared.py
jvoytek, line 1: @ Line 39, sometimes date_str is already a datetime.date `TypeError: strptime() argument 1 must be str, not datetime.date`

python/sidecar/references/service.py
```diff
@@ -0,0 +1,209 @@
...
+    ReferenceCreate
+        The metadata of the reference to create.
+    """
+    response = requests.get(url)
+
+    if response.status_code != 200:
+        raise HTTPError(f"Unable to fetch {url} (status code: {response.status_code})")
+
+    if not metadata.source_filename:
+        metadata.source_filename = f"{metadata.title[:200]}.pdf"
```
cguedes, line 43: Is `metadata.title[:200]` a way of truncating the title to a max of 200 characters? If so, seems too much. The filename should be pretty opaque for the client code (and server) so I think that would be safer to just use an ID (maybe of the reference) instead of the title, that might also include invalid characters for a url.

gjreda: I can see the point that 200 might be too long though, so I'll shorten this to 100 for now.

I had the thought of showing the ID, but since we show the filename in the file browser on the left side, I thought seeing a random UUID would be a confusing user experience.

We also can't rely on the filename from the URL, since it's often a random ID or even sometimes just `pdf`.

IMO it'd be best if we asked a user to provide a filename when adding the reference from search results. I liken this to when you're adding a file to VS Code or downloading a PDF to your local -- you need to provide a name for it. I think a similar principle applies when adding a new reference from a URL. Interaction with the reference in the left pane is going to require knowing its name, so I think the user should provide it.

Issue #361, opened 2023-08-17
Add unit tests for the editor changes


sehyod:
Improve code coverage of custom TipTap extensions

PR #553, opened 2023-09-15
feat: block browsing

### LINKED ISSUE #151: Esc should toggle between block browsing and text editing

sehyod:
- [x] Pressing Escape toggles block browsing
- [x] Pressing Enter while in block browsing mode focuses the content of the current block
- [x] Arrows change the selected block
  - Up: selects the previous sibling or the parent node
  - Down: selects the first child or the next sibling node
  - Left: selects the parent node
  - Right: selects the first child node
- [x] Left Arrow (resp. Right Arrow) collapses (resp. uncollapses) nodes
- [x] Shift+ Up/Down Arrow expands the selection
  - Up: adds the previous sibling block to current selection or selects the parent block
  - Down: adds the next sibling block to current selection
- [x] Cmd+A only selects the content of the current blocks (pressing again selects everything)
- [x] Cmd+A while in block browsing node selects all the blocks
- [x] Removing and copy/pasting block selection works as expected
- [x] Cmd+Shift+ Up/Down Arrows moves the selected blocks
  - Up: moves above collapsed sibling, or indents the selection if it has an uncollapsed sibling, or moves above parent
  - Down: moves after collapsed sibling, or moves inside uncollapsed sibling if there is one, or unindents the selection 
- [x] Tab (resp. Shift+Tab) indent (resp. unindent) the selected blocks
- [x] Formatting buttons work as expected when in block browsing mode
- [x] List buttons work as expected when in block browsing mode (transform the top blocks but not their children)
- [x] Heading buttons are disabled when in block browsing mode (implementing them would require a more logic and I don't think it's worth it since I don't think it is a common use case to transform multiple blocks into heading at the same time)

This PR decreases the coverage, I am planning on adding tests in a follow-up PR (related to #361)

Here is a short video to show some of these features:
https://github.com/refstudio/refstudio/assets/58954208/5f666193-cf47-47ca-8f73-1b9d0d4cda5c

Closes #151


src/features/textEditor/components/tipTapNodes/notionBlock/plugins/blockBrowsingPlugin.ts
```diff
@@ -0,0 +1,393 @@
...
+        console.error('Block browsing is enabled but no block is selected');
+        return;
+      }
+
+      const selectedBlock = view.state.doc.nodeAt(selection.head)!;
+      tr.setMeta(blockBrowsingPluginKey, true);
+      const depthDifference = selection.$from.depth - selection.$to.depth;
+      const endPos = selection.actualTo; // end of selection
+
+      switch (event.code) {
```
cguedes, line 85: nit: I see that this is the core logic for the block navigation. Can you make the code for each case in this switch own function?

cguedes (approved):

cguedes (approved):

Issue #138, opened 2023-06-12
Allow users to search and select references


hammer:
In addition to PDF uploads, it would be nice if users could search and select references from something like the [Semantic Scholar (S2) API](https://api.semanticscholar.org/api-docs/graph).

hammer:
@shauryr This video shows an example of how EndNote allows references to be added from an external reference database: https://youtu.be/S3xo6ZjBV6U?t=40. I'd of course want a better UX for our product, but hopefully you can get the general idea. I'd like to use S2 as our first external source. A lot of this work will be in the UI, I think (cc @cguedes @sehyod)

shauryr:
@hammer thanks for sharing the video. It looks great. 
Semantic Scholar has a `search endpoint` where we can send our query and get the metadata of the paper -
[endpoint link](https://api.semanticscholar.org/graph/v1/paper/search?query=literature+graph&offset=0&limit=10&fields=title,venue,year
)
Here it will return 10 papers with the metadata` title, venue, year` but we can add more keys like `abstract` and `authors` if needed. We can also increase the size of the number of papers returned. If the paper is open access we can `wget` the paper and run grobid for full text extraction. 

I also want to add that the ranking of the search results is custom and is detailed here - [blog by Sergey](https://blog.allenai.org/building-a-better-search-engine-for-semantic-scholar-ea23a0b661e7) ; in practice exact title search is great, but for other queries it is not perfect, especially long queries will usually fail (limitation of there elasticsearch implementation). For our purposes it should be fine, and if we want to improve it we can later look into the SERP API.

shauryr:
@gjreda I've made my first commit! - https://github.com/refstudio/refstudio/commit/b7d0adaa582750af1a80ddc825d26d29c9b6fa2f 
This script enables search functionality using S2 search endpoint. Please review and let me know if there are any improvements or changes you would suggest.

cguedes:
@shauryr have you considered using https://github.com/danielnsilva/semanticscholar instead of RAW http request to the API?

gjreda:
@shauryr Some general thoughts:
1. re: @hammer's comment about this living on the front-end, @cguedes and I think it's worth briefly discussing that on tomorrow's call. I could see it living in either place, but do think type-ahead should live on the front-end given our current architecture.
[...]

PR #557, opened 2023-09-19
138 semantic scholar search

jvoytek:
#138 

- This branch is based off of #548 
- This seems to be mostly working with the exception of PDF's ingested from URL.
- Still needs unit tests but wanted to push so that others can test integration between UI and new endpoint


gjreda:
> This seems to be mostly working with the exception of PDF's ingested from URL.

Checking a handful of these and all the http responses are 403. We should think through a way to communicate to the user that we added the reference, but were unable to fetch the PDF for this (or other) reasons.

src/features/s2search/S2SearchModal.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+  const handleKeyWordsEntered = useCallback(
+    (searchTerm: string, limit?: number) => {
+      setLoading(true);
+      loadSearchResults(searchTerm, limit)
+        .then(() => {
+          setLoading(false);
+        })
+        .catch(() => {
+          console.log('error');
```
sehyod, line 234: Shouldn't this be a `console.error`?

sehyod (commented):
This looks great! I have left a few comments about the frontend.

On a higher level, closing and re-opening the modal resets the state, so if you add a reference, close the modal, open the modal and then search for the same reference again, you are able to add it a second time and you don't have access to the remove button. Maybe that's ok, but I wanted to point it out.

Also, it doesn't seem to be working for references with a PDF in the desktop app: I get an `Upload Error`

Screenshot 2023-09-20 at 10 00 22

src/features/references/editor/ReferencesTableView.tsx
```diff
@@ -157,7 +160,10 @@ export function ReferencesTableView({ defaultFilter = '' }: { defaultFilter?: st
           size="M"
           text="Add"
           type="secondary"
-          onClick={() => emitEvent('refstudio://menu/references/upload')}
+          onClick={(e) => {
+            e.stopPropagation();
+            showUploadContextMenu(e);
```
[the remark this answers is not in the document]
cguedes: I think it's ok for now. Maybe we can have a [split-button-dropdowns](https://getbootstrap.com/docs/4.0/components/dropdowns/#split-button-dropdowns) in the future.

src/events.ts
```diff
@@ -14,6 +14,7 @@ interface RefStudioEvents {
   'refstudio://menu/file/close': undefined;
   'refstudio://menu/file/close/all': undefined;
   'refstudio://menu/settings': undefined;
+  'refstudio://menu/search': undefined;
```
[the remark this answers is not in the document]
cguedes: I agree. We should have this action available under `refstudio://menu/references/search` (alongside with `refstudio://menu/references/upload`).

src/wrappers/contextMenus.css
```diff
@@ -7,9 +7,9 @@ body {
...
 /* This is the class name defined by react-contexify */
 .contexify {
-  @apply z-modals min-w-min p-1;
+  @apply z-modals min-w-min p-0;
 }
 
-.contexify .contexify_itemContent {
-  @apply px-2 py-0.5;
+.contexify_itemContent {
+  @apply p-3;
```
cguedes, line 14: This was made to prevent the "+ Add" popup to look like 

image

And instead like this

image

The side effect of this change is that now all the other context menu items also look the same:

image

Instead of the original

image


This is another reason to reimplement the popup.

tailwind.config.js
```diff
@@ -248,7 +248,7 @@ export default {
       },
     },
   },
-  plugins: [debugPlugin(), autocompleteCustomComponentsPlugin()],
+  plugins: [debugPlugin(), autocompleteCustomComponentsPlugin(),],
```
cguedes, line 251: nit: Remove that extra comma.

cguedes (changes requested):
Overall looks great 👏 

I agree with @sehyod that is odd that we can see that a reference was added (when first added) but when we reopen the modal we can't see that.

We should have a way to know that the reference was added via S2 and then cross-check.

I also think you should reimplement the "+ Add" button dropdown and not use the context menu feature.

cguedes:
@gjreda still in the topic of the PDF file name. I was able to select papers with "invalid" characters for a URL segment (i.e: ? /) as you can see in the following images.

What do you think of doing a simple cleanup replacing all chars not equal to `[0-9A-Za-z ]` to a single dash `-` (or nothing)?

image

image

jvoytek:
> Also, it doesn't seem to be working for references with a PDF in the desktop app: I get an `Upload Error`
> 
> Screenshot 2023-09-20 at 10 00 22

The problem with PDFs in the desktop app is probably the same as for the web. There are still lots of references with a openAccessPdf url that can't be retrieved for different reasons.  Sometimes the new endpoint responds with 200 OK and just ignores the PDF (this shows as "Added" in the interface but the PDF icon won't show in Table or Edit view), but if it encounters an uncaught exception and returns 500 we show the user "Upload Error." There is probably a better user experience for these uncaught exceptions, at least saving the metadata without the PDF.

This is an example of a payload that returns 500:

`{"type":"metadata","url":"https://arxiv.org/pdf/2101.11735","metadata":{"source_filename":"","citation_key":"","doi":"","title":"Wave Dark Matter","abstract":"We review the physics and phenomenology of wave dark matter: a bosonic dark matter candidate lighter than about 30 eV. Such particles have a de Broglie wavelength exceeding the average interparticle separation in a galaxy like the Milky Way and are, thus, well described as a set of classical waves. We outline the particle physics motivations for such particles, including the quantum chromodynamics axion as well as ultralight axion-like particles such as fuzzy dark matter. The wave nature of the dark matter implies a rich phenomenology: ▪  Wave interference gives rise to order unity density fluctuations on de Broglie scale in halos. One manifestation is vortices where the density vanishes and around which the velocity circulates. There is one vortex ring per de Broglie volume on average. ▪  For sufficiently low masses, soliton condensation occurs at centers of halos. The soliton oscillates and undergoes random walks, which is another manifestation of wave interference. The halo and subhalo abundance is expected to be suppressed at small masses, but the precise prediction from numerical wave simulations remains to be determined. ▪  For ultralight ∼10−22 eV dark matter, the wave interference substructures can be probed by tidal streams or gravitational lensing. The signal can be distinguished from that due to subhalos by the dependence on stream orbital radius or image separation. ▪  Axion detection experiments are sensitive to interference substructures for wave dark matter that is moderately light. The stochastic nature of the waves affects the interpretation of experimental constraints and motivates the measurement of correlation functions. Current constraints and open questions, covering detection experiments and cosmological, galactic, and black hole observations, are discussed.","contents":"","published_date":"2021-00-27","authors":[{"full_name":"L. Hui"}]}}`

This one returns 200 (with no PDF):

`{"type":"metadata","url":"http://www.thelancet.com/article/S1474442214703205/pdf","metadata":{"source_filename":"","citation_key":"","doi":"","title":"Hungry and angry: could we survive a zombie apocalypse? \n               \n                  \n                     \n                        Timothy\n                        Verstynen\n                     \n                     \n                        Bradley\n                        Voytek\n             ","abstract":"","contents":"","published_date":"2015-03-1","authors":[{"full_name":"J. Morgan"}]}}`

This one returns 200 (with a PDF on both web and desktop app):

`{"type":"metadata","url":"https://www.biorxiv.org/content/biorxiv/early/2016/10/16/081125.full.pdf","metadata":{"source_filename":"","citation_key":"","doi":"","title":"Inferring synaptic excitation/inhibition balance from field potentials","abstract":"Neural circuits sit in a dynamic balance between excitation (E) and inhibition (I). Fluctuations in this E:I balance have been shown to influence neural computation, working memory, and information processing. While more drastic shifts and aberrant E:I patterns are implicated in numerous neurological and psychiatric disorders, current methods for measuring E:I dynamics require invasive procedures that are difficult to perform in behaving animals, and nearly impossible in humans. This has limited the ability to examine the full impact that E:I shifts have in neural computation and disease. In this study, we develop a computational model to show that E:I ratio can be estimated from the power law exponent (slope) of the electrophysiological power spectrum, and validate this relationship using previously published datasets from two species (rat local field potential and macaque electrocorticography). This simple method--one that can be applied retrospectively to existing data--removes a major hurdle in understanding a currently difficult to measure, yet fundamental, aspect of neural computation.","contents":"","published_date":"2016-09-16","authors":[{"full_name":"Richard Gao"},{"full_name":"E. Peterson"},{"full_name":"Bradley Voytek"}]}}`

@gjreda maybe there are some more edge cases in PDF retrieval we can account for on the back end?

jvoytek:
> This looks great! I have left a few comments about the frontend.
> 
> On a higher level, closing and re-opening the modal resets the state, so if you add a reference, close the modal, open the modal and then search for the same reference again, you are able to add it a second time and you don't have access to the remove button. Maybe that's ok, but I wanted to point it out.

It would be easy enough not to reset the state on close so that you could remove the same reference again when you open it again (without searching again) but I hesitate to allow for removal of the references after a new search has been performed. 

We would need to save the paperId on our end to have any hope of checking for existing references and even then I would be wary of S2's ability to always return the same paperId (I just don't know if or how often those ids change). I think the risk of not being able to add references would be worse than the ability to add multiples of the same reference, but maybe worth a follow up issue...

gjreda:
> We would need to save the paperId on our end to have any hope of checking for existing references and even then I would be wary of S2's ability to always return the same paperId (I just don't know if or how often those ids change)

@jvoytek There's a `metadata` field on the `ReferenceCreate` object that is mostly unused throughout the application. You should be able to pass the `paperId` using this field and we'll save it.

I don't mean to scope creep this PR, but did want to call that out as an option. I don't have a good since of potential problems with `paperId` itself.

src/events.ts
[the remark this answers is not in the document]
jvoytek: Changed the event name and added action to References menu, command palette, and "no references" view in side panel.

gjreda:
image
Something about this particular result causes the Add request to fail. It seems like the request does not get created properly on the frontend.
```
  File "/Users/greg/Dropbox/refstudio/python/sidecar/references/router.py", line 55, in ingest_references
    project_id, metadata=request.metadata, url=request.url
AttributeError: 'IngestUploadsRequest' object has no attribute 'metadata'
```

src/api/s2SearchAPI.ts
```diff
@@ -0,0 +1,84 @@
...
+      const date = new Date(reference.publicationDate);
+      return (
+        date.getFullYear().toString() +
+        '-' +
+        ('0' + date.getMonth().toString()).slice(-2) +
+        '-' +
+        date.getDate().toString()
+      );
+    } else {
+      return reference.year?.toString();
```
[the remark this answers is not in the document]
jvoytek: Jan 1 of that year is all we can do unless we have a way to store only the year

src/api/s2SearchAPI.ts
```diff
@@ -0,0 +1,84 @@
...
+  const pdfRequestBody: IngestMetadataRequest = {
+    type: 'metadata',
+    url: s2SearchResult.openAccessPdf,
+    metadata,
+  };
+
+  const status = await apiPost(
+    '/api/references/{project_id}',
+    { path: { project_id: projectId } },
+    s2SearchResult.openAccessPdf ? pdfRequestBody : genericRequestBody,
```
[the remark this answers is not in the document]
jvoytek: I think you're right. These types were a work in progress when I first implemented this method.

src/components/Button.tsx
```diff
@@ -23,36 +25,46 @@ export function Button({
...
+          'bg-btn-bg-primary-disabled text-btn-txt-primary-disabled': !inheritColor && disabled && type === 'primary',
+          'bg-btn-bg-primary-default': inheritColor && type === 'primary',
         },
         {
-          'bg-btn-bg-secondary-default text-btn-txt-secondary-default': type === 'secondary',
-          'bg-btn-bg-secondary-disabled text-btn-txt-secondary-disabled': disabled && type === 'secondary',
+          'bg-btn-bg-secondary-default text-btn-txt-secondary-default': !inheritColor && type === 'secondary',
+          'bg-btn-bg-secondary-disabled text-btn-txt-secondary-disabled':
+            !inheritColor && disabled && type === 'secondary',
+          'bg-btn-bg-secondary-default': inheritColor && type === 'secondary',
```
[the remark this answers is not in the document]
jvoytek: We need a more robust design system that contains button styles for all the buttons we need. I think this is ok for now but I agree it could be better. inheritColor allows the text color to be pass as className prop.

src/wrappers/contextMenus.css
jvoytek: This was intentional. There doesn't seem to be official "design" for the context menus and with the redesign more white space seemed desirable to me. @hammer what do you think?

jvoytek:
> @gjreda still in the topic of the PDF file name. I was able to select papers with "invalid" characters for a URL segment (i.e: ? /) as you can see in the following images.
> 
> What do you think of doing a simple cleanup replacing all chars not equal to `[0-9A-Za-z ]` to a single dash `-` (or nothing)?
> 
> image image

Single dash seems like a good option to me.

jvoytek:
> > We would need to save the paperId on our end to have any hope of checking for existing references and even then I would be wary of S2's ability to always return the same paperId (I just don't know if or how often those ids change)
> 
> @jvoytek There's a `metadata` field on the `ReferenceCreate` object that is mostly unused throughout the application. You should be able to pass the `paperId` using this field and we'll save it.
> 
> I don't mean to scope creep this PR, but did want to call that out as an option. I don't have a good since of potential problems with `paperId` itself.

I'll work on unit tests for existing functionality and issues with the core functionality then make a new PR for the paperId changes so we can either take them or leave them depending on the side effects.

jvoytek:
> image Something about this particular result causes the Add request to fail. It seems like the request does not get created properly on the frontend.
> ```
>   File "/Users/greg/Dropbox/refstudio/python/sidecar/references/router.py", line 55, in ingest_references
>     project_id, metadata=request.metadata, url=request.url
> AttributeError: 'IngestUploadsRequest' object has no attribute 'metadata'
> ```

I've seen this error a few times and it does look like a problem with the request body but I'm not sure it is (when it happens I usually check the request body to make sure). I just tried that reference specifically and all is working properly (the reference is added with the PDF). I may have inadvertently fixed the problem with other changes or something about the response from S2 changed, or maybe I didn't find the same instance of this references or ?? Can you try it again with the changes I just pushed and/or find another instance of a reference that results in the `AttributeError: 'IngestUploadsRequest' object has no attribute 'metadata'`?

When I tried that reference this was request body:

```json
{
"type":"metadata",
"url":"http://arxiv.org/pdf/2101.04622",
"metadata":{
    "source_filename":"",
    "citation_key":"",
    "doi":"",
    "title":"Communication-safe web programming in TypeScript with routed multiparty session types",
    "abstract":"Modern web programming involves coordinating interactions between browser clients and a server. Typically, the interactions in web-based distributed systems are informally described, making it hard to ensure correctness, especially communication safety, i.e. all endpoints progress without type errors or deadlocks, conforming to a specified protocol. We present STScript, a toolchain that generates TypeScript APIs for communication-safe web development over WebSockets, and RouST, a new session type theory that supports multiparty communications with routing mechanisms. STScript provides developers with TypeScript APIs generated from a communication protocol specification based on RouST. The generated APIs build upon TypeScript concurrency practices, complement the event-driven style of programming in full-stack web development, and are compatible with the Node.js runtime for server-side endpoints and the React.js framework for browser-side endpoints. RouST can express multiparty interactions routed via an intermediate participant. It supports peer-to-peer communication between browser-side endpoints by routing communication via the server in a way that avoids excessive serialisation. RouST guarantees communication safety for endpoint web applications written using STScript APIs. We evaluate the expressiveness of STScript for modern web programming using several production-ready case studies deployed as web applications.",
    "contents":"",
    "published_date":"2021-01-12",
    "authors":[
        {"full_name":"Anson Miu"},
        {"full_name":"Francisco Ferreira"},
        {"full_name":"N. Yoshida"},
        {"full_name":"Fangyi Zhou Imperial College London"},
        {"full_name":"Bloomberg"}
    ]
}
}
```

python/sidecar/references/service.py
```diff
@@ -12,9 +14,51 @@
...
+        if "application/pdf" not in response.headers.get("content-type", ""):
+            reason = f"Unable to fetch {url}, content-type is not PDF"
+            logger.info(reason)
+            return False, reason
+
+        return True, ""
+    except requests.exceptions.RequestException:
+        reason = f"Unable to fetch {url}, request exception"
+        logger.info(reason)
+        return False, reason
```
gjreda, line 56: @jvoytek I added some reasons here that we pass back to the user, but I don't really love them. Maybe we have something more user-friendly on the frontend that suggests they visit the URL and manually add the paper, since that's ultimately what all of these are going to boil down to.

Also happy to change them here if you have suggestions.

gjreda:
> maybe there are some more edge cases in PDF retrieval we can account for on the back end?

Demoing web with the changes from [1bc2f6f](https://github.com/refstudio/refstudio/pull/557/commits/1bc2f6fabbd82e098af11a0e0b4f53a315b9ea46) and I think a lot of the PDF edges have been ironed out, but please let me know if you run into more.

src/wrappers/contextMenus.css
hammer: The difference is marginal to my eye from these images, maybe it would be easier to distinguish on a live app. I’m traveling today so don’t block on me, just pick the easiest path and file an issue to address later.

cguedes: @jvoytek, we can go with your change as we don't have a design for the context menu and we also don't have a button dropdown ATM. But we need to revisit this with the design team in the future.

cguedes:
I've tested this and everything is working as expected for the API responses with the "warnings".

python/sidecar/references/service.py
jvoytek: Added a more verbose tooltip when response contains a message

jvoytek:
@cguedes we good to merge?

yarn.lock
cguedes, line 1: I don't think we should commit this lock file. This updates the version of many dependencies and we can have unexpected behaviour.

cguedes (approved):
LGTM but please remove the changes in `yarn.lock`.

src/atoms/s2SearchState.ts
```diff
@@ -0,0 +1,38 @@
...
+  set(searchResultsAtom, searchResults);
+});
+
+export const clearSearchResultsAtom = atom(null, (_get, set) => {
+  set(searchResultsAtom, []);
+});
+
+export const addS2ReferenceAtom = atom(null, async (_get, set, projectId: string, s2SearchResult: S2SearchResult) => {
+  const status = await postS2Reference(projectId, s2SearchResult);
+  await set(loadReferencesAtom, projectId);
```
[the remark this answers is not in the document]
cguedes: I've added this @jvoytek

cguedes (approved):

yarn.lock
cguedes: @jvoytek I've reset this file with [e51fbb2](https://github.com/refstudio/refstudio/pull/557/commits/e51fbb2ae183f840766f12f1bdae2d17efc945e4)

sehyod (approved):

PR #559, opened 2023-09-20
chore: Log to console and file

### LINKED ISSUE #558: [sidecar] Log to console instead of file
We historically logged to a file since we were entirely desktop based and needed to communicate over stdout.

It's easier to debug if we log to the console, so let's do that.

gjreda:
fixes #558

### DIFF
--- a/python/sidecar/config.py
+++ b/python/sidecar/config.py
@@ -18,9 +18,7 @@ REFERENCES_JSON_PATH = Path(os.path.join(PROJECT_DIR, ".storage", "references.js
 logging.root.setLevel(logging.NOTSET)
 
 logger = logging.getLogger()
-handler = logging.FileHandler(
-    os.path.join(os.environ.get("SIDECAR_LOG_DIR", "/tmp"), "refstudio-sidecar.log")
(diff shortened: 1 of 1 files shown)


python/sidecar/config.py
```diff
@@ -18,9 +18,7 @@
 logging.root.setLevel(logging.NOTSET)
 
 logger = logging.getLogger()
-handler = logging.FileHandler(
-    os.path.join(os.environ.get("SIDECAR_LOG_DIR", "/tmp"), "refstudio-sidecar.log")
-)
+handler = logging.StreamHandler()
```
cguedes, line 21: We spoke about keeping the file logging. Can we have that?

gjreda: @cguedes Is your thought to also include the FastAPI server logs in the file?

cguedes (approved):

python/sidecar/config.py
cguedes: No. Not for now. This is ok 👍 with the current logs being written to the file and and to the console.

PR #571, opened 2023-09-21
feat: add min width to sidebars

sehyod:
This computes the percentage corresponding to 256px to give a min width to sidebar panels

image

In a next release `react-resizable-panels` should add support for using pixels unit directly (cf. https://github.com/bvaughn/react-resizable-panels/pull/176). This PR is a temporary solution.


cguedes (approved):
This is great. Thanks.

PR #572, opened 2023-09-21
fix: Ollama doesn't work on desktop / needs to use async streaming

### LINKED ISSUE #569: Ollama doesn't work at all, I also get an error ('async for' requires an object with aiter method, got ModelResponse)

gjreda:
🚧 This is a work in progress 🚧 

fixes #569 

This seems "right" but responses are no longer streamed (at least not for OpenAI) 🤔


gjreda:
Closing this as it turned into a confusing mess locally. See https://github.com/refstudio/refstudio/pull/573 instead

PR #573, opened 2023-09-22
fix: ollama not working on desktop + make /chat endpoints async

### LINKED ISSUE #569: Ollama doesn't work at all, I also get an error ('async for' requires an object with aiter method, got ModelResponse)

gjreda:
fixes #569 

TODO:
- [x] fix tests


gjreda:
@cguedes I still have a couple tests to fix here, but I think this should resolve the async issues we were seeing on desktop. However, with these changes, web doesn't seem to stream anymore -- it should, but I'm stumped as to why it's not.

IMO, I think it'd make the most sense to release without streaming since we won't be able to support it on desktop anymore. I also have some concerns with Ollama support. While it works, I find it to be terribly slow, to the point of not being useful (though when screen sharing today, I did notice that yours seemed snappier than mine, so maybe it's just an issue for me).

gjreda:
Also feel free to push some commits if you want to try something

sehyod:
> However, with these changes, web doesn't seem to stream anymore

It seems to be working fine on my end on web and on the desktop app (with both OpenAI and Ollama) 🎉 

> While it works, I find it to be terribly slow, to the point of not being useful

I think it's fine: people should expect a local model to be slow. We had an issue in the desktop app, with tauri timing out when calling the non-stream endpoint. However we have just realised that we can call the stream endpoint by using regular fetch instead of tauri's (https://github.com/tauri-apps/tauri/discussions/6613#discussioncomment-5507082). This means that with this PR, everything works as expected!

More generally, it also means we can get rid of the `universalRequest` code to simply use the regular `fetch` everywhere and greatly simplify the code. But this is a large change, that we should not do just before the release.

gjreda:
> We had an issue in the desktop app, with tauri timing out when calling the non-stream endpoint.

Ahhh, _this_ must have been part of the issue I was seeing locally then!

Well all of this is great then. I'll get these tests fixed up so we can merge this.

hammer:
> we can get rid of the universalRequest code to simply use the regular fetch everywhere and greatly simplify the code. But this is a large change, that we should not do just before the release.

File an issue please!

cguedes (approved):

cguedes (approved):
👏

cguedes:
> > we can get rid of the universalRequest code to simply use the regular fetch everywhere and greatly simplify the code. But this is a large change, that we should not do just before the release.
> 
> File an issue please!

Issue created in https://github.com/refstudio/refstudio/issues/576

sehyod:
Great! Everything is working as expected

A few comments about error handling: 
 - we should update the frontend to properly show the error message when an error occurs when trying to rewrite selection
 - I don't if we can catch this, but it would be nice, when a user enters an invalid model, to warn them that the model is invalid (or needs to be installed)

cguedes:
Thanks @sehyod please create follow-up issues for those.

gjreda:
> I don't if we can catch this, but it would be nice, when a user enters an invalid model, to warn them that the model is invalid (or needs to be installed)

This was ultimately why we'd hardcoded `llama2` in the first place -- an implicit decision that we were only going to support that for now, just to limit scope (@cguedes I recall chatting about this when we first added Ollama)

sehyod:
I have created follow-up issues for these 2 comments: #577 and #578
I think it's fine for now anyway, since selecting Ollama automatically sets the model to llama2

Issue #561, opened 2023-09-20
Allow for removing references in S2 modal after close & re-open


jvoytek:
> On a higher level, closing and re-opening the modal resets the state, so if you add a reference, close the modal, open the modal and then search for the same reference again, you are able to add it a second time and you don't have access to the remove button. Maybe that's ok, but I wanted to point it out.

We would need to save the paperId on our end to have any hope of checking for existing references and even then I would be wary of S2's ability to always return the same paperId (I just don't know if or how often those ids change). I think the risk of not being able to add references would be worse than the ability to add multiples of the same reference, but maybe worth a follow up issue...

_Originally posted by @jvoytek in https://github.com/refstudio/refstudio/issues/557#issuecomment-1728055601_

State is reset when the modal closes. This is a larger issue than can be handled in current release.

cguedes:
Agree. I think we need to save the source (type and id) for an ingested reference so that we can know how to dedup for scenarios like this one.

PR #581, opened 2023-09-22
561 s2 paper id dedupe

jvoytek:
#561 
- saves the paperId from S2 search results in the metadata JSON string as {sourceId:[paperId], source:'s2'}
- uses any existing sourceId's to create a lookup table for sourceId->reference.id
- any s2 result that _can be matched_ to an existing reference shows [Added] button by default and can be removed in the modal

### DIFF
--- a/src/atoms/s2SearchState.ts
+++ b/src/atoms/s2SearchState.ts
@@ -4,18 +4,35 @@ import { S2SearchResult } from '../api/api-types';
 import { removeProjectReferences } from '../api/referencesAPI';
 import { getS2ReferencesByKeyword, postS2Reference } from '../api/s2SearchAPI';
 import { refreshFileTreeAtom } from './fileExplorerActions';
-import { loadReferencesAtom } from './referencesState';
+import { getReferencesAtom, loadReferencesAtom } from './referencesState';
 
 // #####################################################################################
 // Internal Atoms
 // #####################################################################################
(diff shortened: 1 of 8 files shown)


src/atoms/s2SearchState.ts
```diff
@@ -4,18 +4,33 @@ import { S2SearchResult } from '../api/api-types';
...
+export const paperIdLookupAtom = atom<Record<string, string>>({});
 
 // #####################################################################################
 // References API
 // #####################################################################################
 
 export const getSearchResultsAtom = atom((get) => Object.values(get(searchResultsAtom)));
+export const getPaperIdLookupAtom = atom((get) => get(paperIdLookupAtom));
+
+export const generatePaperIdLookupAtom = atom(null, (get, set) => {
```
cguedes, line 22: @jvoytek I don't think we want to do this. For now, the s2 modal should only be able to match the references uploaded via the modal.

Also, the sourceId for existing references aren't their ID.

src/features/references/editor/ReferenceDetailsCard.tsx
```diff
@@ -122,12 +122,14 @@ const IconButton = ({ title, icon, onClick }: { title: string; icon: ReactElemen
   </button>
 );
 
+type ReferenceDetailsCardItem = Omit<ReferenceItem, 'metadata'>;
```
cguedes, line 125: I don't see why you need this change.

PR #589, opened 2023-09-26
fix: do not indent blocks when moving them

### LINKED ISSUE #586: Draggable blocks have unexpected indent behaviors
When I select a block in notion and move it up or down, it just moves:

https://github.com/refstudio/refstudio/assets/15233/fb271422-f7d6-436b-b178-156f3c7c7fcb

Ref Studio indents it first, then moves it:

https://github.com/refstudio/refstudio/assets/15233/2b785d16-ecbf-478e-9bbe-c0255fda7e2d

Interestingly Ref Studio does the right thing if the block to be jumped over is collapsible:

https://github.com/refstudio/refstudio/assets/15233/2f4020cc-935d-4e30-b075-fdb4ff2100f9

sehyod:
Fixes #586

This changes the logic when moving a block up/down.
Now the block is indented only when its sibling is a list item, an uncollapsed collapsible block or a regular block with children. Otherwise, the block moves before/after its sibling.

https://github.com/refstudio/refstudio/assets/58954208/976144bf-0363-48dc-8292-ad7144360115

### DIFF
--- a/src/features/textEditor/components/tipTapNodes/notionBlock/plugins/blockBrowsingPlugin.ts
+++ b/src/features/textEditor/components/tipTapNodes/notionBlock/plugins/blockBrowsingPlugin.ts
@@ -232,11 +232,24 @@ function moveBlocksDown({ depthDifference, endPos, selection, tr, dispatch }: Bl
   // If the last selected node has a sibling
   if (indexInParent < parent.childCount - 1) {
     const sibling = parent.child(indexInParent + 1);
-    // If the sibling is not a list item or is a collapsed collapsible block, move selection after the sibling
+    // If the sibling is a list item, an uncollapsed collapsible block or has children, make selected blocks its child
     if (
-      !LIST_ITEM_TYPES.includes(sibling.attrs.type as string) ||
-      (sibling.attrs.type === 'collapsible' && sibling.attrs.collapsed)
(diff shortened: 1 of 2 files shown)


src/features/textEditor/components/tipTapNodes/notionBlock/plugins/blockBrowsingPlugin.ts
```diff
@@ -232,8 +232,24 @@ function moveBlocksDown({ depthDifference, endPos, selection, tr, dispatch }: Bl
...
+      const insertPos = tr.mapping.map(endPos + 1 + sibling.firstChild!.nodeSize);
+      tr.replace(insertPos, insertPos, sliceToMove);
+
+      if (selection.from === selection.anchor) {
+        tr.setSelection(new BlockSelection(tr.doc.resolve(insertPos), tr.doc.resolve(insertPos + selectionRange)));
+      } else {
+        tr.setSelection(new BlockSelection(tr.doc.resolve(insertPos + selectionRange), tr.doc.resolve(insertPos)));
+      }
+    } else {
+      // Otherwise, move selection after the sibling
```
sehyod, line 252: This might look like a lot of changes but that is because I inverted the if and else blocks: it feels more natural to treat the special case where the block is indented first, and then have the default case (block moving before or after its sibling) in the else block

cguedes (approved):

PR #592, opened 2023-09-26
feat: allow one reference per citation

### LINKED ISSUE #587: References behave strangely on navigation and editing
I think we need to revisit our approach to how we allow multiple references, perhaps only entering multiple reference mode when the ";" is manually entered, and changing how the reference is represented in the document.

Some issues I've encountered:
- It's not intuitive that the only way to exit the reference-adding workflow is through the right arrow key. I tried enter, space, tab, even right bracket, and none worked. Also noted by @gjreda at https://github.com/refstudio/refstudio/issues/580
- When moving through text with they keyboard using shortcuts such as "option + right arrow" to move one word at a time, for some reason the cursor skips from the reference all the way to the end of the document. This was quite surprising to me!

https://github.com/refstudio/refstudio/assets/15233/93f14b20-2a97-4b70-ad4e-e97a52951261

- If there is a character immediately following the closed right bracket of a reference, it's impossible to place the cursor in front of the character!

https://github.com/refstudio/refstudio/assets/15233/ebd9a35b-1767-4e69-a8a5-2f1938bf9fcb

- Backspace behavior is weird: when trying to delete a reference that includes some trailing text, there's a moment where hitting backspace actually puts a space back into the reference, then hitting backspace again deletes the reference.

https://github.com/refstudio/refstudio/assets/15233/531e08d6-da42-4b4b-a8c6-393cae805c28

sehyod:
Closes #587 

This is a temporary rollback of references feature to only enable one reference per citation (and no extra text)
Having an inline node causes issues, like the cursor not being able to be on both side of the end position of the node, which leads to some weird behaviors (cf. https://discuss.prosemirror.net/t/discussion-inline-nodes-with-content/496). We need to spend more time defining what we want to do for citations with multiple references.
One solution would be that, when pressing `[`, we add the closing bracket and move the cursor behind it once a reference has been selected. The user could then move the cursor back between the brackets to add more text or other references.


sehyod:
This felt too much of a regression, not being able to have multiple references and text in a citation.
We decided to go with #593 instead

